PR #135, opened 2025-08-20
feat: electrum settings

pwltr:
### Description

- add electrum settings screen
- add (fullscreen) scanner screen
- revert convert SuggestionsManager to singleton
- Split SettingsViewModel into extensions

### Notes

NotificationService doesn't yet use the stored electrum server since it is a separate target. If needed we need to create an app group and share UserDefaults via app group identifier

### Screenshot / Video

https://github.com/user-attachments/assets/19cfcf23-cc2d-45a2-9a09-c46127eb23e9

### DIFF
--- a/Bitkit/ViewModels/Extensions/SettingsViewModel+Electrum.swift
+++ b/Bitkit/ViewModels/Extensions/SettingsViewModel+Electrum.swift
@@ -1,0 +1,179 @@
+import Foundation
+
+// MARK: - Electrum Server Management
+
+extension SettingsViewModel {
+    func loadElectrumSettings() {
+        // Check connection status first
+        let isRunning = lightningService.status?.isRunning == true
+        electrumIsConnected = isRunning
+
+        // Update form with current server (stored or default)
+        let currentServer = electrumConfigService.getCurrentServer()
+        updateForm(with: currentServer)
+    }
+
+    func resetElectrumToDefault() async -> (success: Bool, host: String, port: String, errorMessage: String?) {
+        let defaultServer = electrumConfigService.getDefaultServer()
+        updateForm(with: defaultServer)
+
+        return await connectToElectrumServer()
+    }
+
+    func connectToElectrumServer() async -> (success: Bool, host: String, port: String, errorMessage: String?) {
+        electrumIsLoading = true
+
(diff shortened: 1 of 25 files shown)


Bitkit/ViewModels/Extensions/SettingsViewModel+Electrum.swift
```diff
@@ -0,0 +1,179 @@
...
+            Logger.info("Successfully connected to Electrum server: \(electrumCurrentServer.url)")
+
+            return (success: true, host: host, port: port, errorMessage: nil)
+        } catch {
+            electrumIsConnected = false
+            electrumIsLoading = false
+
+            Logger.error(error, context: "Failed to connect to Electrum server")
+
+            return (success: false, host: host, port: port, errorMessage: nil)
```
ovitrif, line 61: In android we're also reconnecting to the previous connection if the new one failed.
Not sure if it's a must -- maybe that's the main use case of the "Reset to Default" button -- but perhaps an improvement idea for the future ;)

ovitrif (commented):
LGTM 🎉 

**Tested** the new electrum settings screen:
- 🟢 submitting details for localhost server from bitkit-docker
- 🟢 scanning electrum server qr code
- 🟢 validation errors
- 🟠  connection error
  - was expecting app to reconnect to the previously working config (see also code comment)
  - was expecting to see app status icon indicating issues

A few notes:
- really like the slide-in/out-from-bottom/top screen transition for the scanner, we should do the same in Android (cc. @jvsena42)
- more as a reminder, now that electrum connection is fully configurable and the app is aware of the connection status, there are a few enhancements that can be done to the **app status** feature:
  - app status component in navbar can indicate if there are issues with the electrum connection
    - same for app status component in drawer menu
  - app status screen can show electrum status row + navigate to electrum settings onclick

Bitkit/ViewModels/Extensions/SettingsViewModel+Electrum.swift
pwltr: Good idea but I'd rather leave the UX like it is in the current production app. I think this could easily be overlooked as the connection working when it in fact failed at fell back to default.

PR #142, opened 2025-09-04
feat(send): add tags to payment

pwltr:
### Description

- `TagManager` now takes care of `lastUsedTags` and temporary tags for send flow
- add `SendTagScreen`
- apply tags for sent payments

### Screenshot / Video

https://github.com/user-attachments/assets/7c89a402-ca54-44b0-abbe-e9ea97b07346


Bitkit/ViewModels/ActivityListViewModel.swift
```diff
@@ -244,9 +201,25 @@ class ActivityListViewModel: ObservableObject {
...
-        saveRecentlyUsedTags()
+    func findActivityAndAddTags(paymentHashOrTxId: String, tags: [String]) async throws {
+        guard !tags.isEmpty else { return }
+
+        // Find the activity by payment ID
+        let activity = try await tryNTimes(
+            toTry: { try await findActivity(byPaymentId: paymentHashOrTxId) },
+            times: 12,
+            interval: 5
+        )
```
jvsena42, line 212: non-blocking: could do a sync if not found on the first attempt
Reference: https://github.com/synonymdev/bitkit-android/blob/b0ba9423d1458941d573bc34174c0086cb389334/app/src/main/java/to/bitkit/repositories/ActivityRepo.kt#L136-L138

jvsena42 (approved):
Found an issue, but can be fixed in other PR
- ⚠ Send -> Add tags -> boost a transaction -> the tags are not being added  to the boosted transaction

pwltr:
> Found an issue, but can be fixed in other PR
> 
> * ⚠ Send -> Add tags -> boost a transaction -> the tags are not being added  to the boosted transaction

Yes I think this will need to be reworked to work similarly to the RN app

PR #178, opened 2025-10-03
fix: disable scroll on widgets edit mode

### LINKED ISSUE #172: Cannot reorder widgets
Observed on IPhone13+

Once you move into widgets editing mode and try to re-order, the whole screen moves when you try to drag a widget.

jvsena42:
### Description

This PR fix the widget reordering issue

### Linked Issues/Tasks
Close #172


pwltr (approved):
Not the best UX if we have to block scrolling but this will be a non issue with the new design anyway

PR #206, opened 2025-11-06
Add VSS backup for all categories

ben-kaufman:
This PR adds support for wallet, metadata, blocktank, activity, and lightning connections categories for VSS backups, and fixes issues from the previous VSS backup PR.

https://github.com/user-attachments/assets/5b327a27-bc6b-4938-a03a-2cbe510eb06f

### DIFF
--- a/Bitkit/Services/LightningService.swift
+++ b/Bitkit/Services/LightningService.swift
@@ -242,9 +242,15 @@ class LightningService {
         }
         Logger.info("LDK synced")
 
-        // Emit state change with sync timestamp
-        let syncTimestamp = UInt64(Date().timeIntervalSince1970)
-        syncStatusChangedSubject.send(syncTimestamp)
+        // Emit state change with sync timestamp from node status
+        let nodeStatus = node.status()
+        if let latestSyncTimestamp = nodeStatus.latestLightningWalletSyncTimestamp {
+            let syncTimestamp = UInt64(latestSyncTimestamp)
+            syncStatusChangedSubject.send(syncTimestamp)
+        } else {
+            let syncTimestamp = UInt64(Date().timeIntervalSince1970)
+            syncStatusChangedSubject.send(syncTimestamp)
+        }
     }
 
     func newAddress() async throws -> String {
--- a/Bitkit/ViewModels/SettingsViewModel.swift
+++ b/Bitkit/ViewModels/SettingsViewModel.swift
@@ -159,27 +159,6 @@ class SettingsViewModel: NSObject, ObservableObject {
         updatePinEnabledState()
     }
 
-    init(
-        lightningService: LightningService = .shared,
-        electrumConfigService: ElectrumConfigService = ElectrumConfigService(),
-        rgsConfigService: RgsConfigService = RgsConfigService()
-    ) {
-        self.lightningService = lightningService
-        self.electrumConfigService = electrumConfigService
-        self.rgsConfigService = rgsConfigService
-
-        // Initialize electrumCurrentServer with current server (stored or default)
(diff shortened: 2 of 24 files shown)


Bitkit/Services/BackupService.swift
```diff
@@ -381,8 +482,95 @@ class BackupService {
...
+            var tagMetadata: [TagMetadataItem] = []
+            do {
+                let activities = try await CoreService.shared.activity.get()
+                for activity in activities {
+                    let activityId = switch activity {
+                    case let .lightning(ln): ln.id
+                    case let .onchain(on): on.id
+                    }
+
+                    let tags = try await CoreService.shared.activity.tags(forActivity: activityId)
```
ovitrif: aside:
I suppose iOS doesn't support tags for receive transactions?!

Not sure it's doable without storing `tagMetadata` separately, and then attaching it to the activity item when the payment for that bip21 is received. 🤔

Bitkit/Models/SettingsBackupConfig.swift
```diff
@@ -0,0 +1,90 @@
...
+        "hasSeenTransferToSpendingIntro",
+        "hasSeenTransferToSavingsIntro",
+        "hasSeenWidgetsIntro",
+        "showHomeViewEmptyState",
+        "appUpdateIgnoreTimestamp",
+        "backupIgnoreTimestamp",
+        "highBalanceIgnoreCount",
+        "highBalanceIgnoreTimestamp",
+        "dismissedSuggestions",
+        "lastUsedTags",
```
jvsena42: these keys already exist in the app, probably with another name

Bitkit/Services/LightningService.swift
```diff
@@ -234,6 +241,10 @@ class LightningService {
             // try? self.setMaxDustHtlcExposureForCurrentChannels()
         }
         Logger.info("LDK synced")
+
+        // Emit state change with sync timestamp
+        let syncTimestamp = UInt64(Date().timeIntervalSince1970)
+        syncStatusChangedSubject.send(syncTimestamp)
```
ovitrif: This ain't bad but it would be better to read the value from `node.status()`, ie `NodeStatus.latestLightningWalletSyncTimestamp` (which is in SECONDS! not millis, so it needs to be multiplied)

ovitrif:
nit: Keeping the logging a bit consistent between the 2 platforms would help us be able to reuse our debugging skills in PROD from iOS to Android.

What I try to say is that the logging that got removed in this PR but were added in the previous one, were fine to remain IMHO.

ovitrif (dismissed):
Tested 🟢 

- Transfer back to savings without 6 block confirmations
- Backup & restore
- **expect** transfer pending balance

ovitrif:
@ben-kaufman I approved the PR because I consider it a valid step forward which doesn't break stuff; but I would still expect the relevant comments to be resolved, either in next PR or here 🙏🏻

Bitkit/ViewModels/SettingsViewModel.swift
```diff
@@ -86,8 +121,43 @@ class SettingsViewModel: ObservableObject {
...
+
     // MARK: - Initialization
 
+    override private init() {
+        lightningService = .shared
+        electrumConfigService = ElectrumConfigService()
+        rgsConfigService = RgsConfigService()
+        electrumCurrentServer = electrumConfigService.getCurrentServer()
+
+        super.init()
```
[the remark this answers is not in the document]
ovitrif: Ah, nvm, I think you fixed it by ensuring SettingsViewModel is always accessed via `.shared`.

PR #208, opened 2025-11-06
Update tags backup and improve logs

ben-kaufman:
Updates the tags to use the BitkitCore functions for upsert and read tags with metadata.
Also updates the logs to match more closely the logs on the android app.


jvsena42 (commented):
https://github.com/user-attachments/assets/19103070-d7f5-4311-a22e-b6b31281c032

The code looks correct, but the app hasn't recovered the tags

ben-kaufman:
@jvsena42 thanks, should be fixed now, tested and it recovered properly.

PR #211, opened 2025-11-07
Vss backup fixes

ben-kaufman:
Add latest backup status and fix backup status issues.

### DIFF
--- a/Bitkit/Services/BackupService.swift
+++ b/Bitkit/Services/BackupService.swift
@@ -31,7 +31,10 @@ class BackupService {
 
     var backupStatusesPublisher: AnyPublisher<[BackupCategory: BackupItemStatus], Never> {
         backupStatusesSubject
-            .removeDuplicates()
+            .removeDuplicates { old, new in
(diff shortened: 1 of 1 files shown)


Bitkit/Resources/Localization/ca.lproj/Localizable.strings
```diff
@@ -260,7 +260,7 @@
 "security__mnemonic_keep_header" = "Manteniu-ho segur";
 "security__mnemonic_data_header" = "Dades de la cartera";
 "security__mnemonic_multiple_header" = "Múltiples dispositius";
-"security__mnemonic_latest_backup" = "<bold>Data última còpia de seguretat:</bold> {time}";
+"security__mnemonic_latest_backup" = "<accent>Data última còpia de seguretat:</accent> {time}";
```
jvsena42: Have updated on Trasifex?

ben-kaufman: No since we don't yet have dedicated Trasifex for the native app and I don't want to affect the RN app. I think we should wait to have the native Trasifex to update no?

PR #219, opened 2025-11-13
Integrate pre activity metadata and receive tags

### LINKED ISSUE #197: Cannot add tag on editing receive invoice
Test: `@onchain_2`
**Steps to reproduce:**
1. Have a new wallet
2. Hit Receive > Edit
3. Try to add tag

**Expected:**
After adding a tag it should be visible on the Edit sheet.

**Actial:**
Tag is not visible after adding.

https://github.com/user-attachments/assets/a1493a1b-6ddf-4988-9d2c-5c15ea1ea6ee

ben-kaufman:
This PR integrates pre-activity metadata from Bitkit Core.

Fixes #197 

The pre-activity metadata is created before the send/ receive activities are created, and allow Bitkit Core to automatically add tags to an activity once created (for sent txs, the previous logic to apply tags was removed). Also for sent txs, it let's Bitkit Core auto fill the address for the activity, and for receive txs it helps in a more efficient lookup for the address the activity belongs to.

This PR also extracts certain UI components related to tags and reuses them across the app.
It also fixes the VSS backup to correctly backup and restore activities tags and pre-activity metadata.

https://github.com/user-attachments/assets/7a962759-251d-466e-b220-2064365821c8

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -238,10 +238,10 @@ jobs:
           cd docker && docker compose down -v || true
 
   e2e-status:
+    if: always() && github.event.pull_request.draft == false
     name: e2e-status
     runs-on: ubuntu-latest
     needs: [e2e-tests]
-    if: always()
     steps:
       - name: Verify all E2E shards succeeded
         run: |


Bitkit/Services/CoreService.swift
```diff
@@ -235,6 +246,118 @@ class ActivityService {
...
+        // Try receiving addresses first, then change addresses
+        if let address = try await searchAddresses(isChange: false) {
+            return address
+        }
+        if let address = try await searchAddresses(isChange: true) {
+            return address
+        }
+
+        // Fallback: return first output address
+        return txDetails.vout.first?.scriptpubkey_address ?? "todo_find_address"
```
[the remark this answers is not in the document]
ben-kaufman: It is what we had before. In theory it should never happen, do you have any suggestions for a better fallback?

jvsena42:
Roadmap item
https://docs.google.com/spreadsheets/d/1NBtx1y-_ZcODo8L0LsbudBNV_9x7cF6MyckR_W1xxq4/edit?pli=1&gid=0#gid=0&range=C14

Bitkit/Views/Wallets/Receive/ReceiveSheet.swift
```diff
@@ -45,8 +46,11 @@ struct ReceiveSheet: View {
         .onAppear {
             wallet.invoiceAmountSats = 0
             wallet.invoiceNote = ""
+            tagManager.clearSelectedTags()
             Task {
                 try? await wallet.refreshBip21(forceRefreshBolt11: true)
+                // Create pre-activity metadata after refreshing the invoice/address
+                await wallet.persistPreActivityMetadata()
```
jvsena42: could delete data with more than 48 hours or with expired invoices, because this method is called multiple times

ben-kaufman: I don't think we should really. If a payment is received then the pre activity is deleted automatically in bitkit core, and if it's not received then it will either be replaced once the user reopens the receive tab, or the user might still be waiting for a payment. So either way I don't see why we should delete after 48 hours.

ben-kaufman:
Issue should now be fixed @jvsena42 

Testing instructions are as you did + click in savings balance -> transfer to spending -> make sure the resulting activity is shown as transfer.

Bitkit/Models/BackupPayloads.swift
```diff
@@ -12,7 +12,7 @@ struct WalletBackupV1: Codable {
 struct MetadataBackupV1: Codable {
     let version: Int
     let createdAt: UInt64
-    let tagMetadata: [ActivityTagsMetadata]
+    let tagMetadata: [PreActivityMetadata]
```
ovitrif: should we name it `preActivityMetadata`? I can do the same in Android ;)

Or even `activityMetadata`, since we won't likely have a different activity metadata soon 🤷🏻

ovitrif:
> Issue should now be fixed @jvsena42
> 
> Testing instructions are as you did + click in savings balance -> transfer to spending -> make sure the resulting activity is shown as transfer.

You mean on the previous PR that was closed in favour of this?!

jvsena42 (approved):
**Lightning**

- Create invoice -> add tags and description -> receive -> check if tags and description were added ✅
- scan invoice -> add tags -> send -> check tags ✅

**On-chain**

- Create BIP 21 invoice -> add tags -> receive -> activity details -> check tags ✅
- Scan BIP 21 invoice -> add tags -> send -> activity details -> check address, fee, federate and tags ✅
- Transfer to spending ✅
- Reset and restore ✅

OBS: On transfer to spending, the pending transfer activities are not being displayed. They are only displayed after confirmed. (can be fixed in #226 )


https://github.com/user-attachments/assets/4d56153a-ebf1-4c85-80f3-d0bfccd9c4a6

ovitrif (commented):
### Tests
1. Receive Tags: all tests ok 🟢 
2. Transfer: test ok 🟢 
3. Send Tags: all tests ok 🟢 

---

While testing receive tags, In found an issue not added in these changes, but related to the feature.

Receive > Edit > Tap Note textfield > **Can't add tags anymore**
Really need to show QR code and then edit again.

PR #230, opened 2025-11-17
Mark RBFed txs as removed from mempool

ben-kaufman:
This PR marks RBFed transactions as removed from mempool.

Testing: Make an onchain transaction, then go to its activity details and boost it.

Simulator Screenshot - iPhone 16 - 2025-11-17 at 12 34 16
Simulator Screenshot - iPhone 16 - 2025-11-17 at 12 34 19

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -107,7 +107,7 @@ jobs:
           # - { name: onchain_boost_receive_widgets, grep: "@onchain|@boost|@receive|@widgets" }
           # - { name: settings, grep: "@settings" }
           # - { name: security, grep: "@security" }
-          - { name: e2e, grep: "@onboarding|@onchain_1|@numberpad|@widgets|@boost|@receive" }
+          - { name: e2e, grep: '^(?!.*@settings_10)(@onboarding|@onchain_1|@numberpad|@widgets|@boost|@receive|@settings)' }
 
     name: e2e-tests - ${{ matrix.shard.name }}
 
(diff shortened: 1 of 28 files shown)


ben-kaufman:
Need to update the e2e test to match the behaviour of not deleting the boosted activities.

ovitrif (approved):
ACK ✅ 

Test: Send onchain → Boost 🟢 

For e2e test to pass pass, please check with @piotr-iohk tomorrow 🙏🏻

piotr-iohk:
> ACK ✅
> 
> Test: Send onchain → Boost 🟢
> 
> For e2e test to pass pass, please check with @piotr-iohk tomorrow 🙏🏻

Yes, it's the `@boost_2 - Can do RBF` test failing as it's not expecting the additional tx which is "Removed from mempool".

Question: Where is this change coming from? I haven't found it reflected in the Figma design, and it seems inconsistent with the current behavior in the Bitkit RN app, where the original transaction is not shown once it is RBF-ed - the wallet only displays the replacement transaction.

I’m also a bit concerned that "Removed from mempool" may feel alarming or unclear for non-technical users. If we want to expose this state, we might want to consider alternative wording or a different UX approach.

jvsena42 (approved):

ben-kaufman:
@piotr-iohk the reason for the label is that eventually we'd want also the other cases of dropped transactions to have that same label.

But we could rename it for this case if that's what we decide.

ovitrif:
@piotr-iohk 

Indeed, for the case when a tx is removed due to RBF (non-technically said: replaced), the current copy (wording) is not the most optimal, I agree.

Though as @ben-kaufman pointed out, the reason we went with this is because "Transaction Removed from Mempool" is also for another case, when a tx might get dropped because it never gets confirmed, or due to a reorg, or if the tx is not propagated enough. (sharing this from DM chats with Ben).

We can use a copy along the lines that you're suggesting:
> "Replaced by higher fee transaction"

I would also like to invite @aldertnl for review & input of this text 🙏🏻 .

Issue #246, opened 2025-11-27
[Bug]: SettingsViewModel has stale state after app wipe


pwltr:
See comment: https://github.com/synonymdev/bitkit-ios/pull/206/files#r2568108660

- [ ] Ensure that the settings view model state is fully reset on app wipe

PR #258, opened 2025-12-02
feat(ui): add balance action toasts

pwltr:
### Description

- add toast when switching currency for the first time
- add toast when hiding balance for the first time

### Screenshot / Video

https://github.com/user-attachments/assets/81fc856e-6cf7-4ed1-969a-7e4260438c40

### DIFF
--- a/Bitkit/Components/MoneyStack.swift
+++ b/Bitkit/Components/MoneyStack.swift
@@ -8,6 +8,7 @@ struct MoneyStack: View {
     var showEyeIcon: Bool = false
     var enableSwipeGesture: Bool = false
 
+    @EnvironmentObject var app: AppViewModel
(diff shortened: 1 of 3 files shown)


pwltr:
> @pwltr there seems to be an issue introduced. "Hide balance on open" setting is not preserved.
> 
>  Simulator.Screen.Recording.-.iPhone.17.-.2025-12-03.at.10.24.09.mov

Hmm, works for me. Could be related to this: https://github.com/synonymdev/bitkit-ios/issues/246 I don't think it is caused by this PR. If you can please retest on master branch.

https://github.com/user-attachments/assets/ce5b92d0-366e-4c25-9e37-be09694df7c1

piotr-iohk (approved):

PR #263, opened 2025-12-04
Update ldk node version to 0.7.0

ben-kaufman:
Updates LDK node version to 0.7.0 

Also updates LDK node config to count unconfirmed incoming balances as spendable

### DIFF
--- a/Bitkit/Services/LightningService.swift
+++ b/Bitkit/Services/LightningService.swift
@@ -46,6 +46,7 @@ class LightningService {
             trustedPeersNoReserve: Env.trustedLnPeers.map(\.nodeId),
             perChannelReserveSats: 1
         )
+        config.includeUntrustedPendingInSpendable = true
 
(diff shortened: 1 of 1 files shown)


ovitrif (approved):
tACK

Tested send unconfirmed funds, right after new wallet + fund via dev settings > blocktank regtest, without mining.

Found a bug in iOS, not related to this though. I tested the same flow and it works in Android.

https://github.com/user-attachments/assets/9f55fe1b-f6a6-438b-a817-ba76c9c2e7c2

Should be fixed separately, not in scope of this PR.

piotr-iohk:
@ben-kaufman passing now. Sorry, it is due to the divergence on e2e tests related to onchain events not merged yet on Android. Therefore for the time being I need to create specific branch on e2e (based of https://github.com/synonymdev/bitkit-e2e-tests/pull/56) for each iOS change.

PR #265, opened 2025-12-05
refactor: move LSP balance calculations to BitkitCore

ben-kaufman:
Refactors channel liquidity calculations from native app to bitkit-core and improves the spending flow limits and UX.

https://github.com/user-attachments/assets/91bee5fb-9d49-4a43-b467-4cd92dca1b4f

### DIFF
--- a/Bitkit/Views/Transfer/SpendingAdvancedView.swift
+++ b/Bitkit/Views/Transfer/SpendingAdvancedView.swift
@@ -102,7 +102,6 @@ struct SpendingAdvancedView: View {
                 blocktankInfo: blocktank.info
             )
 
-            amountViewModel.updateFromSats(transfer.transferValues.defaultLspBalance, currency: currency)
             updateFeeEstimate()
         }
         .onChange(of: lspBalance) { _ in
--- a/Bitkit/Views/Transfer/SpendingAmount.swift
+++ b/Bitkit/Views/Transfer/SpendingAmount.swift
@@ -10,6 +10,7 @@ struct SpendingAmount: View {
 
     @StateObject private var amountViewModel = AmountInputViewModel()
     @State private var isLoading = false
+    @State private var availableAmount: UInt64?
     @State private var maxTransferAmount: UInt64?
 
     private var amountSats: UInt64 {
@@ -17,8 +18,7 @@ struct SpendingAmount: View {
     }
 
     private var isValidAmount: Bool {
-        guard amountSats > 0 else { return false }
-        guard let max = maxTransferAmount, max > 0 else { return false }
+        guard let max = maxTransferAmount else { return false }
         return amountSats <= max
     }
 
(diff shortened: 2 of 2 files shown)


Bitkit/Views/Transfer/SpendingAmount.swift
```diff
@@ -10,20 +10,16 @@ struct SpendingAmount: View {
...
-    // Calculate the maximum amount that can be transferred
-    private var availableAmount: UInt64 {
-        return maxSendableAmount ?? UInt64(wallet.spendableOnchainBalanceSats)
-    }
-
-    private var transferValues: TransferValues {
-        transfer.calculateTransferValues(clientBalanceSat: amountSats, blocktankInfo: blocktank.info)
+    private var isValidAmount: Bool {
+        guard amountSats > 0 else { return false }
+        guard let max = maxTransferAmount, max > 0 else { return false }
```
pwltr: Available amount should be the total onchain balance minus the network fee, like before

Bitkit/Views/Transfer/SpendingAdvancedView.swift
```diff
@@ -99,10 +102,15 @@ struct SpendingAdvancedView: View {
                 blocktankInfo: blocktank.info
             )
 
+            amountViewModel.updateFromSats(transfer.transferValues.defaultLspBalance, currency: currency)
```
pwltr: Initial balance should be 0 on this screen

Bitkit/Views/Transfer/SpendingAmount.swift
ben-kaufman: But the the available and max amount mismatch

Bitkit/Views/Transfer/SpendingAmount.swift
```diff
@@ -10,20 +10,16 @@ struct SpendingAmount: View {
...
 
-    // Calculate the maximum amount that can be transferred
-    private var availableAmount: UInt64 {
-        return maxSendableAmount ?? UInt64(wallet.spendableOnchainBalanceSats)
-    }
-
-    private var transferValues: TransferValues {
-        transfer.calculateTransferValues(clientBalanceSat: amountSats, blocktankInfo: blocktank.info)
+    private var isValidAmount: Bool {
+        guard amountSats > 0 else { return false }
```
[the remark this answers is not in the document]
pwltr: I think the reasoning was that users may want to purchase inbound liquidity only

Bitkit/Views/Transfer/SpendingAdvancedView.swift
pwltr: IMO if the user navigates to this screen they intend to set a custom amount, defaulting to the amount that was already set is forcing them to delete via keyboard first. I would prefer if we don't make UX changes without discussion with design team first, even small ones.

Bitkit/Views/Transfer/SpendingAmount.swift
ben-kaufman: So the issue is before, the Max could be more than the available, which really didn't make sense I think (the Figma shows the opposite case which is ok)

PR #290, opened 2025-12-23
feat: support RN app migration

ben-kaufman:
This PR adds support for migrating an existing React Native Bitkit into the Bitkit native app.
The migration passes:
- Balances
- Channels
- Activities
- Settings
- Widgets
- Tags and notes
- Closed Channels

https://github.com/user-attachments/assets/44ec4b26-f89d-4e72-a0c0-50cf44459a64

What's still needed:
1. New design for the loading screen and success toast of migration
2. Support for restoring a wallet created in RN app which was later deleted. Ie. create wallet in RN, use it, delete the app, install native, recover using mnemonic.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,8 +21,16 @@ This app integrates with:
 ```bash
 # Standard build - Open Bitkit.xcodeproj in Xcode and build
 
-# E2E test build (uses local Electrum backend)
+# E2E test build (uses local Electrum backend by default)
 xcodebuild -workspace Bitkit.xcodeproj/project.xcworkspace \
+  -scheme Bitkit \
+  -configuration Debug \
+  SWIFT_ACTIVE_COMPILATION_CONDITIONS='$(inherited) E2E_BUILD' \
+  build
+
+# E2E test build with network Electrum and regtest (Info.plist build setting)
+E2E_BACKEND=network E2E_NETWORK=regtest \
+  xcodebuild -workspace Bitkit.xcodeproj/project.xcworkspace \
   -scheme Bitkit \
   -configuration Debug \
   SWIFT_ACTIVE_COMPILATION_CONDITIONS='$(inherited) E2E_BUILD' \
@@ -197,7 +205,7 @@ New feature (`TransferTrackingManager`) tracks pending transfers to handle edge 
 - The app currently runs on **regtest only** (see `LightningService.swift:92` guard)
 - VSS (Versioned Storage Service) authentication is not yet implemented
 - Electrum/Esplora server URLs are configurable via `Env`
-- E2E builds use local Electrum backend via `E2E_BUILD` compilation flag
+- E2E builds use local Electrum backend via `E2E_BUILD` compilation flag (override with `E2E_BACKEND`/`E2E_NETWORK` build settings)
 
 ### Error Handling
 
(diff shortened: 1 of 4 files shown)


Bitkit/Services/MigrationsService.swift
```diff
@@ -1,101 +1,1110 @@
...
+import Security
 
-typealias Expression = SQLite.Expression
+// MARK: - MMKV Parser
 
-class MigrationsService {
+/// Lightweight parser for MMKV binary format (react-native-mmkv)
+/// MMKV stores data as: [4-byte size (little-endian)][4-byte marker][key-value pairs...]
+/// Each pair: [varint key_length][key_bytes][varint value_length][value_bytes]
+struct MMKVParser {
```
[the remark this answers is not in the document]
ben-kaufman: Haven't considered it initially but I think it's not really worth it especially since it's just for a one time migration

Bitkit/AppScene.swift
```diff
@@ -127,6 +133,32 @@ struct AppScene: View {
...
+            ProgressView()
+                .scaleEffect(1.5)
+                .tint(.white)
+
+            VStack(spacing: 8) {
+                Text("Updating Wallet")
+                    .font(.system(size: 24, weight: .semibold))
+                    .foregroundColor(.white)
+
+                Text("Please wait while we update the app...")
```
jvsena42: nit: could also say "Don't close the app"

ben-kaufman: I think we can just wait until design is provided before we update this, it's practically a placeholder for now

Bitkit/Services/CoreService.swift
```diff
@@ -154,6 +154,42 @@ class ActivityService {
...
+                    isSeen = onchain.seenAt != nil
+                case let .lightning(lightning):
+                    id = lightning.id
+                    isSeen = lightning.seenAt != nil
+                }
+
+                if !isSeen {
+                    try await ServiceQueue.background(.core) {
+                        try BitkitCore.markActivityAsSeen(activityId: id, seenAt: timestamp)
+                    }
```
jvsena42: nit: maybe could apply some parallelization in case the wallet has many activities, but I'm not sure if it would make a big difference

jvsena42:
It is not recovering the lighting channels and transactions. Tried twice

1. Install RN app
2. open first channel
3. Mine 1 block
4. do lighting transactions
5. open second channel
6. mine 1 block
7. Install native app
8. Message recover with success
9. Lighting data not recovered

Mnemonics:
`vocal farm title vendor plug rain strong idle rally kidney front tone
`
[logs.txt](https://github.com/user-attachments/files/24367328/logs.txt)

ben-kaufman:
Just retried with the seed you shared, when I restored in the RN channels were forced closed so I opened new ones, then installed the native app and channels and LN txs passed with no issue. Can you please recheck the bundle ID matches that of RN and the network in both is set to regtest?

jvsena42 (dismissed):
Tested on #292 :

1. On RN, open a channel
2. Perform lightnign and on-chain transaction
3. On native: Remove "-regtest" from Bitkit and BitkitNotification Bundle identifier
4. Install app
5. All activities and channels were restored with success

PR #296, opened 2025-12-30
chore: update bundle id for release

pwltr:
### Description

Updates bundle id and version for production release.

Also aligns default network for build configuration with the react native app:
- Regtest for debug builds (local development and testing)
- Mainnet for release builds (production)

Build configuration can be changed in either Xcode or by using Sweetpad command `Select build configuration`.


pwltr:
@piotr-iohk Could you update the bundleId in the e2e setup to "to.bitkit"?

To https://github.com/synonymdev/bitkit-e2e-tests/issues/85, this PR does not do much in that regard except differentiate between debug -> regtest and release -> mainnet. Is that good enough for now or do we need more configurability?

PR #303, opened 2026-01-07
Migrate to-do cards

ben-kaufman:
Migrates the todo cards in the suggestions below balance to hide the dismissed ones after RN app migration.

Note: iOS has Get Paid and Shop which Android and React Native don't have and we should likely fix this to match.


pwltr (approved):
tested, looks good

Issue #200, opened 2025-11-03
Coin selection failed when sending near max amount from the wallet


piotr-iohk:
Test: `@onchain_3  - Avoids creating a dust output and instead adds it to the fee`
**Steps to reproduce:**
1. Have a wallet with, say `100 000 000`
2. Try to send available max - `300` (`99 999 888 - 300 = 99 999 588`)

**Expected:**
The transaction should be completed.

**Actual:**
Send Error - Coin selection failed

https://github.com/user-attachments/assets/62505bfc-35fd-4534-941b-7a162aa25b5a

PR #318, opened 2026-01-11
fix: VSS race condition causing memory corruption during wallet restore

jvsena42:
### Description

Problem
Memory corruption crash during wallet restore:

`malloc: Incorrect checksum for freed object 0x1013da8d8: probably modified after being freed.`
The crash occurs when multiple concurrent tasks call VssBackupClient.awaitSetup() simultaneously.

Root Cause
The awaitSetup() method has a race condition. When multiple tasks call it concurrently:

Task A checks if let existingSetup = isSetup → nil → proceeds to create setup task
Task B checks if let existingSetup = isSetup → nil (Task A hasn't assigned yet) → also proceeds
Both tasks create separate setupTask instances
Both call setup() → double initialization of Rust FFI → memory corruption

The logs show VSS setup being called 7 times concurrently:

```
DEBUG: VSS client setting up… - VssBackupClient [VssBackupClient.swift: setup(walletIndex:) line: 18]
```

Solution:
Use a Swift actor to ensure thread-safe setup coordination. The actor isolates the setupTask state, ensuring only one setup runs at a time.

### Linked Issues/Tasks

N/A

### Screenshot / Video

To test:

- Restore a wallet
- Should complete without crash
- Check logs show  "VSS client setting up" twice:
  - Initial setup when the app starts the restore flow
  - Setup after reset() is called before performing the full restore (to ensure a clean state)

### DIFF
--- a/Bitkit/Services/LightningService.swift
+++ b/Bitkit/Services/LightningService.swift
@@ -14,6 +14,8 @@ class LightningService {
 
     private var channelCache: [String: ChannelDetails] = [:]
 
+    private var storedEventCallback: ((Event) -> Void)?
+
     var syncStatusChangedPublisher: AnyPublisher<UInt64, Never> {
         syncStatusChangedSubject.eraseToAnyPublisher()
     }
@@ -187,7 +189,11 @@ class LightningService {
             throw AppError(serviceError: .nodeNotSetup)
         }
 
-        listenForEvents(onEvent: onEvent)
+        if let onEvent {
+            storedEventCallback = onEvent
+        }
+
+        listenForEvents(onEvent: storedEventCallback)
 
         Logger.debug("Starting node...")
         try await ServiceQueue.background(.ldk) {
@@ -214,7 +220,7 @@ class LightningService {
         }
     }
 
-    func stop() async throws {
+    func stop(clearEventCallback: Bool = false) async throws {
         defer {
             // Always try to unlock, even if stopping fails
             try? StateLocker.unlock(.lightning)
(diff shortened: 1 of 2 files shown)


Bitkit/Services/BackupService.swift
```diff
@@ -504,8 +504,6 @@ class BackupService {
 
     func getLatestBackupTime() async -> UInt64? {
         do {
-            try await vssBackupClient.setup()
```
jvsena42: `getRemoteBackupTimestamp()` eventually calls `VssBackupClient.shared.getObject()`, which internally calls `awaitSetup()`

ovitrif:
Makes perfect sense to use actor here. The reason I didn't in Android is because Kotlin doesn't have it at all and the `CompletableDefferred` is quite nice in there to solve for lack of actor model in its concurrency.

But the proper way to mirror that code in Swift is likely with actor.

jvsena42:
OBS: The crash only happened to me one time, couldn't reproduce it again

ovitrif (commented):
utAck

Testing (sanity migration check)

pwltr (approved):
Tested, looks good

ovitrif:
@jvsena42 what are "integration-tests" and why they keep failing? 🤔

jvsena42:
> @jvsena42 what are "integration-tests" and why they keep failing? 🤔

Are tests that depend on iOS environment or an API call, like blocktank
The failing test is related to coin selection, probably related to #200

ovitrif:
YAY! CI PASSED 🚀 🚀 🚀

PR #319, opened 2026-01-11
fix: keychain persistence

### LINKED ISSUE #293: App is persisting keychain data after uninstall
The app Groups and Keychain DO persist between app uninstalls by default on iOS simulators and sometimes on physical devices. This is the likely cause:

- Keychain data (mnemonics, PINs) persists across reinstalls unless explicitly deleted
- App Group containers may not be fully cleaned up on uninstall

Solution:

Modify [Keychain.swift:31](vscode-webview://1gupj4drib073smbpmrn08u9qlfsgs92taje7rmefbtf3pfce3js/Bitkit/Utilities/Keychain.swift#L31) to use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly`:

`kSecAttrAccessible as String: kSecAttrAccessibleWhenUnlockedThisDeviceOnly as String`

This change means:
✅ Keychain data will NOT persist after app deletion
✅ Data is deleted when the app is uninstalled
❌ Data will NOT transfer to new devices via iCloud Keychain
❌ Data will NOT survive device backups/restores

jvsena42:
Fix #293 

## Description

This PR prevents sensitive keychain data (mnemonics, passphrases, PINs) from being recovered after app uninstall by:

1. Encrypting all keychain entries with AES-GCM using a key stored in the app's file system.
2. Detecting orphaned keychain data on fresh install and wiping it automatically.
3. Using a Documents marker file to reliably detect app reinstallation.

> Breaking Change ⚠️
Users upgrading from master to this branch will have their wallet wiped and will need to restore from their recovery phrase.

This only affects the upgrade path. Normal usage after installation and migration from RN **works correctly.**

**Why Documents Marker?**

The encryption key is stored in App Group storage, which persists after uninstall on iOS. The Documents directory is reliably deleted on uninstall, so the marker file serves as the reliable detection mechanism.

## Testing Scenarios

1. Fresh install → Creates wallet normally with encrypted keychain.
2. Uninstall + reinstall → Detects orphaned data, wipes, shows onboarding.
3. App wipe (Settings) → Cleans everything including encryption key.
4. RN migration → Migrates data correctly, then encrypts.
5. Upgrade from master → Wipes wallet (expected, see **Breaking Change** above).

https://github.com/user-attachments/assets/a1d3a2a5-a887-4658-9fd5-e64eebedd8e2

https://github.com/user-attachments/assets/9dbab146-5a98-43a2-89e8-b3f8aaef91d9

https://github.com/user-attachments/assets/473f39f3-9941-4383-9904-8766b5794777

https://github.com/user-attachments/assets/7ca0ae47-823f-4bd8-bf2f-631b24835354


jvsena42:
A simpler approach is rely only on Documents Marker to identify orphaned keychain data (RN does this), but encrypt the data makes more difficult to restore the keychain data by accident

jvsena42:
> when testing migration got this:
> 
> DEBUG: Encrypted data (73 bytes → 101 bytes) - KeychainCrypto [KeychainCrypto.swift: encrypt(_:) line: 78] ERROR❌: Decryption failed: The operation couldn’t be completed. (CryptoKit.CryptoKitError error 3.) - KeychainCrypto [KeychainCrypto.swift: decrypt(_:) line: 104] ERROR❌: Failed to decrypt bip39_mnemonic_0: decryptionFailed - Keychain [Keychain.swift: load(key:) line: 148] PERF: get(filter:txType:tags:search:minDate:maxDate:limit:sortDirection:) took 0.0 seconds on core queue [ServiceQueue.swift: background(_:_:functionName:) line: 58] ERROR❌: RN migration failed: failedToDecrypt - AppScene [AppScene.swift: checkAndPerformRNMigration() line: 354] ERROR❌: Decryption failed: The operation couldn’t be completed. (CryptoKit.CryptoKitError error 3.) - KeychainCrypto [KeychainCrypto.swift: decrypt(_:) line: 104] ERROR❌: Failed to decrypt bip39_mnemonic_0: decryptionFailed - Keychain [Keychain.swift: load(key:) line: 148] PERF: get(filter:txType:tags:search:minDate:maxDate:limit:sortDirection:) took 0.0 seconds on core queue [ServiceQueue.swift: background(_:_:functionName:) line: 58]
> 
> and an error toast and the app got stuck on the splash screen

This happens when install from master -> uninstall -> install RN -> migrate to this branch

In Keychain.save(), encrypt() was called first (creating the encryption key), then load() was called to check if data exists. But load() now tried to decrypt the old plaintext data with the newly created key, causing decryption failure.

I'm considering implementing an even simpler solution to avoid such edge cases

PR #337, opened 2026-01-15
fix: migrate activities for unsupported addresses

ben-kaufman:
Fixes missing activities when migrating RN app which had activities of unsupported address types.

Testing:

Could use this seed which has example recceving to each address type this making 2 send txs:
amazing shy syrup aspect mistake hungry about sail predict rifle lawsuit supply

Import this seed into the RN app, or create a new wallet and generated activity for different address types.
Run the native app to trigger migration and make sure all activities pass correctly.

Note: The RN app doesn't back up on chain activities on the remote backup, which means if user recovers using mnemonic and RN backup activities from unsupported address types won't show up.

### DIFF
--- a/Bitkit/Services/CoreService.swift
+++ b/Bitkit/Services/CoreService.swift
@@ -338,7 +338,12 @@ class ActivityService {
         guard case let .onchain(txid, _) = payment.kind else { return }
 
         let paymentTimestamp = payment.latestUpdateTimestamp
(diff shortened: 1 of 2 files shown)


jvsena42:
Testing...

ovitrif (approved):
tAck

https://github.com/user-attachments/assets/57e19fba-8277-48cc-bbf7-28422599774e

jvsena42:
tAck

PR #364, opened 2026-01-16
Basic bug report template

piotr-iohk:
### Description

Basic for now. Can be expanded later if needed.


piotr-iohk:
CI not needed for this one.

Issue #363, opened 2026-01-16
[Bug]: Spending balance suddenly 0 and cannot see my connection


catch-21:
I unlocked my iphone and suddenly my spending balance is 0. I have no lightning connections (open or closed). Restarting bitkit doesn't recover it. I should have a few thousand in spending balance. As seen in the video, I have recently sent some bitcoin via LN.

[bitkit_logs_2026-01-16_15-52-55.zip](https://github.com/user-attachments/files/24675584/bitkit_logs_2026-01-16_15-52-55.zip)

https://github.com/user-attachments/assets/4a940cb0-8683-4e98-89ce-16d3c83b6794

catch-21:
Looks like ldk initiated force close, see 'bogus' in my logs
```
% grep -r bogus
./ldk_foreground_2026-01-16_15-43-00.log:2026-01-16 15:43:03.679 DEBUG [lightning::ln::channelmanager:11736] Sending bogus ChannelReestablish for unknown channel 3fd8b3cd6f696f42582e5f6caf049390cebd9ad145037e544592d254fad544a6 to force channel closure
./ldk_foreground_2026-01-16_15-43-00.log:2026-01-16 15:43:03.683 DEBUG [lightning::ln::channelmanager:11736] Sending bogus ChannelReestablish for unknown channel 6a83507e6d95a6f25243b193b4f26e27f00ff79ab0066e8d8e7697dbddefba1a to force channel closure
./ldk_foreground_2026-01-16_15-44-49.log:2026-01-16 15:44:54.199 DEBUG [lightning::ln::channelmanager:11736] Sending bogus ChannelReestablish for unknown channel 3fd8b3cd6f696f42582e5f6caf049390cebd9ad145037e544592d254fad544a6 to force channel closure
./ldk_foreground_2026-01-16_15-44-49.log:2026-01-16 15:44:54.200 DEBUG [lightning::ln::channelmanager:11736] Sending bogus ChannelReestablish for unknown channel 6a83507e6d95a6f25243b193b4f26e27f00ff79ab0066e8d8e7697dbddefba1a to force channel closure
```

I had 2 channels.

PR #376, opened 2026-01-20
fix(restore): clear VSS store ID cache on wipe

pwltr:
### Description

- clear VSS store ID cache
    - on app wipe (probably all that's needed)
    - when restoring from a react native backup (additional defensive code)
- also defensively normalizes empty passphrase
    - when visiting "Restore" screen, tapping "Advanced" and not entering a passphrase it is set to empty string
    - under the hood it is normalized I'm fairly sure but this ensures that

Related: https://github.com/synonymdev/bitkit-ios/issues/363


ben-kaufman (approved):
utAck

PR #381, opened 2026-01-21
fix: migration when user has connection issues

ben-kaufman:
Fixes infinite load screen and adds retry when back online logic for migration when user has connection issues/ is offline

### DIFF
--- a/Bitkit.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Bitkit.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -7,7 +7,7 @@
       "location" : "https://github.com/synonymdev/bitkit-core",
       "state" : {
         "branch" : "master",
-        "revision" : "31ebadcbc1fcbfcdd6e1f66e28b6208bc64a55f8"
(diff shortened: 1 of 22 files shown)


piotr-iohk:
Retested:

**Migration with normal network**
 - started migration
 - no migration loader screen, Bitkit load screen for a moment and went directly too Homescreen
 - activity tags not migrated in activity list search by tags (although assigned to activities)
logs: [bitkit_logs_2026-01-22_12-20-31.zip](https://github.com/user-attachments/files/24794346/bitkit_logs_2026-01-22_12-20-31.zip)

https://github.com/user-attachments/assets/983fbf6d-49af-464b-b84b-a8f341cbb939

**Migration with 100% loss network**
 - situation pretty much the same, no migration loader screen, went directly to Homescreen, but offcourse nothing got migrated due to bad network
 - after bringing back the network and restarting the app all seemed to migrated ok though
 
logs: 
[bitkit_logs_2026-01-22_12-34-00.zip](https://github.com/user-attachments/files/24794574/bitkit_logs_2026-01-22_12-34-00.zip)

https://github.com/user-attachments/assets/adb6602b-aa35-41b2-8b1b-1cd26cd07d03

ovitrif:
@pwltr @piotr-iohk @jvsena42
Someone of you still available to (re) review & test this today or I should do it instead?!

PR #416, opened 2026-01-28
fix: handle dust change in transfers and sends

ben-kaufman:
### Description

This PR fixes accurate full-balance sending for Blocktank channel purchases:

1. **Two-pass fee estimation** in SpendingAmount ensures LSP fee calculation matches actual order creation (fixes off-by-few-sats errors where max showed higher than actually sendable)
2. **Dust detection** in SpendingConfirm uses sendAll when change would be dust, avoiding unspendable outputs
3. **sendAll support** for normal on-chain sends when change would be below dust limit
4. **Safety validation** ensures max sendable amount hasn't changed due to fee rate changes before executing transfer

### Technical Details

The core issue was that `lspBalance` values depend on `clientBalanceSat`, so fee estimation with `maxSendable` produced different fees than actual order creation with `clientBalance`. The two-pass estimation now:
1. First pass: estimates with `maxSendable` to get approximate `clientBalance`
2. Second pass: recalculates `lspBalance` using `approxClientBalance` to match what order creation will use

### Linked Issues/Tasks

Fixes MAX transfer to spending failing with insufficient funds error when calculated change would be dust.

#### MAX Transfer to Spending
1. Have on-chain balance (e.g., 5000 sats)
2. Go to Transfer > To Spending
3. Press MAX button
4. Tap Continue - amounts should add up correctly on confirm screen
5. Swipe to transfer - should succeed without "Insufficient funds" error

#### Near-MAX Normal Send
1. Have on-chain balance
2. Scan/enter an on-chain address
3. Enter an amount that leaves change below dust limit (~547 sats)
4. Confirm screen should show and swipe should work
5. Transaction should succeed

### DIFF
--- a/Bitkit/Utilities/Errors.swift
+++ b/Bitkit/Utilities/Errors.swift
@@ -1,7 +1,7 @@
 import Foundation
 import LDKNode
 
-enum CustomServiceError: Error {
+enum CustomServiceError: LocalizedError {
     case nodeNotSetup
     case nodeNotStarted
     case onchainWalletNotInitialized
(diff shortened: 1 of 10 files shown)


piotr-iohk:
The exact case from #200 is still failing for me here giving "Coin selection error". Interestingly it passes on Android. 
The cases below work fine here. `Near-MAX Normal Send` - works fine for 100k sats, the #200 case uses `100 000 000` sats and then it fails with "Coin selection error":

> #### MAX Transfer to Spending
> 1. Have on-chain balance (e.g., 5000 sats)
> 2. Go to Transfer > To Spending
> 3. Press MAX button
> 4. Tap Continue - amounts should add up correctly on confirm screen
> 5. Swipe to transfer - should succeed without "Insufficient funds" error
> 
> #### Near-MAX Normal Send
> 1. Have on-chain balance
> 2. Scan/enter an on-chain address
> 3. Enter an amount that leaves change below dust limit (~547 sats)
> 4. Confirm screen should show and swipe should work
> 5. Transaction should succeed

Note the `MAX Transfer to Spending` case fails on Android, created issue for it: https://github.com/synonymdev/bitkit-android/issues/745

PR #424, opened 2026-02-02
fix: add retry logic for channel monitor retrieval during migration

jvsena42:
### Description

This PR fixes a critical bug where channel monitor retrieval failures during migration from the React Native app were silently dropped, potentially causing permanent loss of Lightning channel state.

1. Adds `retrieveChannelMonitorWithRetry` method with configurable retry attempts and linear backoff
2. Implements parallel retrieval with proper error tracking for each channel
3. Logs individual failures with channel IDs for debugging
4. Validates expected vs retrieved monitor count with warnings

This is the iOS equivalent of the Android fix: https://github.com/synonymdev/bitkit-android/pull/760

### Linked Issues/Tasks

Related: https://github.com/synonymdev/bitkit-android/pull/760

### Screenshot / Video

N/A - internal migration logic fix

### DIFF
--- a/Bitkit/Services/MigrationsService.swift
+++ b/Bitkit/Services/MigrationsService.swift
@@ -1927,7 +1927,7 @@ extension MigrationsService {
                 if attempt < maxAttempts {
                     let delayMs = baseDelayMs * UInt64(attempt)
                     Logger.debug(
-                        "Retrying channel monitor retrieval for \(channelId) (attempt \(attempt + 1)/\(maxAttempts)) after \(delayMs)ms",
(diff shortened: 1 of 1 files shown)


jvsena42:
> Is it intentional that we continue with restoring if one or more channel monitors fail to retrieve? Seems to me like we should only restore if everything goes right, and if not show the restore error.

I thought that if an error happens for some reason that we don't know yet, other than just connection, the user could be stuck outside the wallet

On the case that we got from support, the wallet migrate almost all the channels except one, that was force closed

But I agree, the restore error is also a valid solution, since the user could also try to restore from mnemonics if they are backed up

pwltr (approved):
utACK, comments above not applicable to this PR.

PR #451, opened 2026-02-19
Use sendAll when change would be dust instead of creating dust change outputs

ben-kaufman:
## Problem
When sending close to the maximum amount, the app sometimes created dust change outputs instead of using sendAll. This happened because the dust check used the sendAll fee (1-output tx) instead of the normal fee (recipient + change). That overestimated expected change and made the dust check pass when it should not.

## Solution
1. Correct fee in dust check
Use the normal fee (recipient + change outputs) when deciding if change would be dust. If change is below the dust limit, use sendAll so the extra amount goes to fees instead of a dust output.
2. DustChangeHelper
Added DustChangeHelper.shouldUseSendAllToAvoidDust() to centralize the logic.
Added unit tests for the dust-change decision.
Applied in both SendConfirmationView (onchain send) and SpendingConfirm (Blocktank channel purchase).

### DIFF
--- a/Bitkit/Utilities/DustChangeHelper.swift
+++ b/Bitkit/Utilities/DustChangeHelper.swift
@@ -1,0 +1,21 @@
+import Foundation
+
+/// Helper for determining when to use sendAll to avoid creating dust change outputs.
+enum DustChangeHelper {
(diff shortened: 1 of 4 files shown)


jvsena42 (approved):
Ack

ovitrif (approved):
utAck

Issue #448, opened 2026-02-18
[Feat]: Send: add pending payment UI


pwltr:
### Problem or use case

If a payment is being held by the payee, the UI just stays in the loading state forever.

### Proposed solution

After 10 seconds, show a pending payment UI that the user can close

<img width="950" height="1600" alt="Image" src="https://github.com/user-attachments/assets/8c14fbe0-fb0d-4302-940c-3dff7dae471a" />

PR #465, opened 2026-02-26
feat(send): add pending payment UI

pwltr:
### Description

For hold invoices (or slow payments), when the send doesn’t complete in 10 seconds, show a pending screen and, when the payment later succeeds or fails, show a toast and navigate to Success or Failure.

### Linked Issues/Tasks

Closes https://github.com/synonymdev/bitkit-ios/issues/448

### Screenshot / Video

https://github.com/user-attachments/assets/f6d5bff0-6ff0-4571-8dc5-f38e94b582a6

### QA Notes

Use `bitkit-docker` to generate and settle/cancel hold invoices

```
    holdinvoice [amount] [-m memo]     Create a hold invoice
    settleinvoice <preimage>           Reveal a preimage and use it to settle the corresponding invoice
    cancelinvoice <payment_hash>       Cancels a currently open invoice
```

### DIFF
--- a/Bitkit/ViewModels/AppViewModel.swift
+++ b/Bitkit/ViewModels/AppViewModel.swift
@@ -44,10 +44,10 @@ class AppViewModel: ObservableObject {
     @AppStorage("hasSeenTransferToSavingsIntro") var hasSeenTransferToSavingsIntro: Bool = false
     @AppStorage("hasSeenWidgetsIntro") var hasSeenWidgetsIntro: Bool = false
 
-    // When to show empty state UI
+    /// When to show empty state UI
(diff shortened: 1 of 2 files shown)


piotr-iohk:
A nit perhaps. In case we close the "Pending payment" screen and use Bitkit while waiting on tx, we do not get any notification after the tx is resolved or fails, the status of tx just changes "silently". I wonder if we'd like to have a notification in such a case:

Successful tx:

https://github.com/user-attachments/assets/8d7f64f5-f132-4e5a-9aaa-00d075228c78

Failed tx:

https://github.com/user-attachments/assets/05f22537-f19e-404a-bb12-43b758fc662f

piotr-iohk (approved):
Retested. LGTM o/

PR #469, opened 2026-03-02
Fix: Show savings activity list when balance is zero but activities exist

### LINKED ISSUE #464: ios: savings with zero balance opens send-to-savings screen instead of activity
## What happened?
When savings balance is `0`, tapping the savings card opens the generic screen with the message **\"SEND BITCOIN TO YOUR SAVING BALANCE\"** instead of showing the savings activity list.

## Expected behavior
Tapping savings should open the savings activity list view, even when savings balance is `0`.

## Steps to Reproduce
1. Have on-chain funds.
2. Transfer all funds to savings.
3. Tap on savings.
4. Observe generic \"SEND BITCOIN TO YOUR SAVING BALANCE\" screen instead of savings activity list.

## Logs / Screenshots / Recordings

https://github.com/user-attachments/assets/f4cf7f1a-5bcd-4f85-bde8-86ff08b41bfa

## Bitkit Version
2.0.5

## Device / OS
iOS sim

## Reproducibility
Always

## Additional context
- Related context: #411, #461

ben-kaufman:
## Summary

Fixes #464 — When savings balance is `0`, tapping the savings card showed the generic "SEND BITCOIN TO YOUR SAVING BALANCE" empty state screen instead of the savings activity list.

## Root Cause

In `SavingsWalletView`, the activity list (ScrollView with ActivityList) was only rendered when `wallet.totalOnchainSats > 0`. When balance reached zero (e.g., after transferring all funds to spending), the view fell through to the `EmptyStateView` overlay, hiding all transaction history.

## Changes

- **`SavingsWalletView.swift`**: Show activity list when balance > 0 OR when there are onchain activities. Only show empty state when both balance is zero AND no activities exist. Transfer button remains gated behind positive balance.
- **`SpendingWalletView.swift`**: Applied the same fix for consistency — show activity list when balance > 0 OR when there are lightning activities.

## Testing

- Have on-chain funds
- Transfer all funds from savings to spending
- Tap on savings card → should show the activity list with past transactions
- Fresh wallet with no history → should still show the empty state screen
- Same behavior verified for spending wallet view

### DIFF
--- a/Bitkit/Views/Wallets/SavingsWalletView.swift
+++ b/Bitkit/Views/Wallets/SavingsWalletView.swift
@@ -6,6 +6,12 @@ struct SavingsWalletView: View {
     @EnvironmentObject var navigation: NavigationViewModel
     @EnvironmentObject var wallet: WalletViewModel
 
+    /// Whether there are any onchain activities to display
+    private var hasOnchainActivities: Bool {
+        guard let activities = activity.onchainActivities else { return false }
+        return !activities.isEmpty
+    }
+
     /// Calculate remaining duration for force close transfers
     private var forceCloseRemainingDuration: String? {
         guard let claimableAtHeight = wallet.forceCloseClaimableAtHeight,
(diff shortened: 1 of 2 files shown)


pwltr (approved):

ovitrif:
> @ovitrif on android it seems correct already actually.

Thx. Idk home page then has this issue?!

After send all to another address, it shouldn't show the banner but the activity imho.

@aldertnl

PR #488, opened 2026-03-12
feat: send flow design updates v59

pwltr:
### Description

This PR updates the send flow to conform with the new design update v59.

- add camera permission UI
- add button to hide/show details on confirm screen
- add ability to switch payment method (savings/spending) on confirm screen
- change custom fee rate to default to fee rate set in transaction speed settings, fall back to slow
- show relative date time for invoice expiry
- add image animation for swiper
- refactoring and cleanup

### Screenshot / Video

Simulator Screenshot - iPhone 17 - 2026-03-12 at 12 52 00

https://github.com/user-attachments/assets/49b4902e-4565-4aa0-a371-66d984acb84d

https://github.com/user-attachments/assets/a72405d3-6a0e-4d8b-90c1-04a228420874

### DIFF
--- a/Bitkit/Managers/CameraManager.swift
+++ b/Bitkit/Managers/CameraManager.swift
@@ -8,7 +8,7 @@ final class CameraManager {
 
     var hasPermission: Bool = false
 
-    init() {
+    private init() {
         refreshPermission()
     }
 
--- a/Bitkit/ViewModels/AppViewModel.swift
+++ b/Bitkit/ViewModels/AppViewModel.swift
@@ -577,7 +577,6 @@ extension AppViewModel {
         selectedWalletToPayFrom = .onchain // Reset to default
         lnurlPayData = nil
         lnurlWithdrawData = nil
-        resetManualEntryInput()
     }
 }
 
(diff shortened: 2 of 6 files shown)


pwltr:
> > > It was also "camera permissions alert" that was blocking (should be handled now). The other thing is that the alert (if not accepted or denied) stays displayed even if the app gets terminated, doesn't look like this would be expected?
> > > Screen.Recording.2026-03-23.at.17.40.04.mov
> > 
> > 
> > For me it doesn't let me terminate the app while this dialog is showing on the simulator or device. Maybe something e2e specific? Is the app crashing or why is it terminated?
> 
> Yes, it doesn't let the app terminate. (In the recording I terminated it with `xcrun simctl terminate booted to.bitkit`). It _might_ affect e2e tests in a way if certain test misses hitting `Allow` (or somehow fails on that screen), then all the subsequent tests are blocked by this window, even though new instance of the app is started every time.

There is not much we can do to prevent this, this dialog is entirely managed by the operating system, I guess that is also why it stays there even if the app that called the API is terminated. Do you think it is better to skip asking for camera permission in e2e tests?

pwltr:
@piotr-iohk Still not sure why e2e tests are failing. If there is anything I should try let me know.

Otherwise this is ready for review.

Issue #423, opened 2026-02-01
[Bug]: Swipe to go back not working


pwltr:
### Expected behavior

Usually it is possible to swipe from the left of the viewport to navigate back.

### Bitkit Version

2.0.3

### Device / OS

iPhone SE / 26.1

### Reproducibility

None

PR #491, opened 2026-03-13
fix(ui): re-enable swipe-to-back gesture

pwltr:
### Description

We are using a custom `NavigationBar` and hiding the system bar with `.navigationBarHidden(true)` on each screen. In SwiftUI, that hides the UINavigationController’s navigation bar. The interactive pop gesture (swipe from the left edge) is managed by that same navigation controller, and when the bar is hidden the system disables that gesture. This re-enables the interactive pop gesture by configuring the underlying `UINavigationController` and allows the gesture when there is more than one view controller on the stack, except when explicitly disabled with by using `.allowSwipeBack(false)`

### Linked Issues/Tasks

Closes https://github.com/synonymdev/bitkit-ios/issues/423

### Screenshot / Video

https://github.com/user-attachments/assets/f800f3ba-5d66-4b4f-8f25-ec7061865649

### QA Notes

- check for quirks/freezes
- should only be able to navigate back on screens that have a back button

### DIFF
--- a/Bitkit/Extensions/UINavigationController+SwipeBack.swift
+++ b/Bitkit/Extensions/UINavigationController+SwipeBack.swift
@@ -1,0 +1,25 @@
+import UIKit
+
+/// Shared state for swipe-back gesture. Views can set this via the `.allowSwipeBack(_:)` modifier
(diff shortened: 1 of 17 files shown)


ovitrif (approved):
tAck

was missing this so much, thx! 🙌🏻

PR #480, opened 2026-03-06
fix(lightning): fix orphaned channels migration

pwltr:
### Description

Don't overwrite channel manager, we only need the monitors to sweep the funds from force closed channels

### Screenshot / Video

https://github.com/user-attachments/assets/15494924-032e-47d9-8ff9-030503dc2a54

[logs.txt](https://github.com/user-attachments/files/25791346/logs.txt)

### Repro steps

1. Install RN app (master)
2. Open channel
3. Backup seed
4. Delete RN app
5. Install native app (modified: fail to retrieve channel monitor & disable orphaned channel recovery)
6. Restore from seed (channel is lost w/o sweep)
7. Open new channel
8. Update app (master)
9. new channel force closed

PR #501, opened 2026-03-18
feat: one-time stale channel monitor recovery (v2)

jvsena42:
### Description

This PR adds automatic one-time recovery for users affected by stale channel monitors caused by the RN migration overwrite bug (#462, #495).

On `BuildError.ReadFailed`, the app automatically retries the LDK node build once with `accept_stale_channel_monitors` enabled. A persisted `staleMonitorRecoveryAttempted` UserDefaults flag ensures this only happens once — the flag is set on any successful build, so unaffected users see zero impact.

This is a clean cherry-pick of the recovery logic from #500 onto `release-2.1.1`, without the unrelated pubky work that was in the original branch.

### Linked Issues/Tasks

- Supersedes #500 (same fix, cleaner branch base)
- Related: #462, #480, #495
- Depends on: [synonymdev/ldk-node#76](https://github.com/synonymdev/ldk-node/pull/76)

### Screenshot / Video

N/A - backend logic change

### QA Notes

1. Reproduce the stale monitor state (overwrite a channel monitor in VSS with an older update_id)
2. Launch the app — first build fails with ReadFailed
3. Verify the retry succeeds and the node starts
4. Check logs for "Stale monitor recovery: build succeeded with accept_stale"
5. Kill and relaunch — verify normal startup (no retry triggered)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Bitkit/Services/LightningService.swift
+++ b/Bitkit/Services/LightningService.swift
@@ -9,6 +9,17 @@ class LightningService {
     private var node: Node?
     var currentWalletIndex: Int = 0
 
+    // MARK: - Stale monitor recovery (one-time recovery for channel monitor desync)
+
+    private static let staleMonitorRecoveryAttemptedKey = "staleMonitorRecoveryAttempted"
+
+    /// Whether we've already attempted stale monitor recovery (prevents infinite retry).
+    /// Persisted so the retry only happens once, even across app restarts.
+    private static var staleMonitorRecoveryAttempted: Bool {
+        get { UserDefaults.standard.bool(forKey: staleMonitorRecoveryAttemptedKey) }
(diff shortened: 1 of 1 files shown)


Bitkit/Services/LightningService.swift
```diff
@@ -124,22 +135,59 @@ class LightningService {
...
+                        storeId: storeId,
+                        fixedHeaders: [:]
+                    )
+                }
+            } catch let error as BuildError {
+                guard case .ReadFailed = error, !Self.staleMonitorRecoveryAttempted else {
+                    throw error
+                }
+
+                // Build failed with ReadFailed — likely a stale ChannelMonitor (DangerousValue).
```
jvsena42: Replace with specific exception

PR #503, opened 2026-03-19
chore: use e2e updater url for e2e builds

piotr-iohk:
This PR points E2E builds to a separate updater release.json hosted in bitkit-e2e-tests, so production critical updates don't block E2E test runs.

### Description

The production updater currently has `critical: true` for both platforms. E2E builds from master with lower build numbers hit the blocking update screen immediately, making all tests fail.

When `isE2E` is true, `updaterUrl` now resolves to:
`https://github.com/synonymdev/bitkit-e2e-tests/releases/download/updater/release.json`

That file has `buildNumber: 0, critical: false` by default, so E2E builds are never blocked. When update flow testing is needed, the E2E release.json can be updated independently.

### Linked Issues/Tasks

N/A

### Screenshot / Video

N/A

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Bitkit/Constants/Env.swift
+++ b/Bitkit/Constants/Env.swift
@@ -277,7 +277,12 @@ enum Env {
     static let playStoreUrl = "https://play.google.com/store/apps/details?id=to.bitkit"
     static let githubUrl = "https://www.github.com/synonymdev/bitkit"
     static let githubReleasesUrl = "https://www.github.com/synonymdev/bitkit/releases"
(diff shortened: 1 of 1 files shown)


piotr-iohk:
> Maybe better to get into the habit of making sure build number updates make it to master before release (and consequent updating of release.json). This is an annoyance in development as well.

I believe they do make it to master before release.json update. But if particular branch is not updated then there is an issue.
(Also if you want to build some older versions to e.g. test migration)

PR #526, opened 2026-04-21
fix: replace Zip dependency with native ZipService

pwltr:
### Description

Switch log zipping to an internal NSFileCoordinator-based service, add safer filename/overwrite/temp-file handling, and cover key zip edge cases with tests.

### QA Notes

Zipping logs works as before with same compression level.

1. Go to Settings -> Advanced -> Lightning Connections
2. Tap "Export Logs"
3. Unpack the .zip and verify log files are usable


piotr-iohk (approved):
LGTM.

I also noticed exported log ZIPs now go to the iOS app container’s `tmp/` folder instead of `Documents` / the container root. That still seems fine for the share sheet, including “Save to Files”, since the file is just the temporary source that gets copied to the selected destination.

PR #529, opened 2026-04-24
fix(lightning): update scorer params

pwltr:
### Description

Update scorer params to improve payment success rate.


ovitrif:
can be removed for now.

PR #530, opened 2026-04-28
[codex] Add Pubky Ring auth callbacks

ben-kaufman:
## Summary
- Add Pubky Ring `x-success`, `x-cancel`, `x-error`, and `x-source=Bitkit` callback parameters to generated auth URLs.
- Handle Pubky Ring return callbacks through the existing Bitkit deeplink path.
- Reset Ring auth waiting/loading UI on cancel or error callbacks while keeping relay completion as the source of truth for successful auth.
- Add focused tests for callback URL generation and callback parsing.

## Why
Pubky Ring now supports custom callback URLs. Adding these parameters lets Ring return users to Bitkit after approval, cancellation, or error instead of leaving the app switch flow hanging.

## Validation
- `git diff --check`
- `swiftformat` was run on the touched Swift files.
- Full `xcodebuild` was attempted, but it is blocked by the existing external `VssRustClientFfi` / `Uniffi` generated bindings issue.

### DIFF
--- a/Bitkit/MainNavView.swift
+++ b/Bitkit/MainNavView.swift
@@ -247,9 +247,9 @@ struct MainNavView: View {
                 Logger.info("Received deeplink: \(url.absoluteString)")
 
                 if let callback = PubkyRingAuthCallback.parse(url: url) {
-                    await pubkyProfile.handleAuthCallback(callback)
+                    let didHandleCallback = await pubkyProfile.handleAuthCallback(callback)
 
-                    if case let .error(message) = callback {
+                    if didHandleCallback, case let .error(message, _) = callback {
                         app.toast(
                             type: .error,
                             title: t("profile__auth_error_title"),
--- a/Bitkit/MainNavView.swift
+++ b/Bitkit/MainNavView.swift
@@ -246,6 +246,20 @@ struct MainNavView: View {
             Task {
                 Logger.info("Received deeplink: \(url.absoluteString)")
 
+                if let callback = PubkyRingAuthCallback.parse(url: url) {
+                    await pubkyProfile.handleAuthCallback(callback)
+
+                    if case let .error(message) = callback {
+                        app.toast(
+                            type: .error,
+                            title: t("profile__auth_error_title"),
+                            description: message ?? t("other__qr_error_text")
+                        )
+                    }
+
+                    return
+                }
+
                 do {
                     try await app.handleScannedData(url.absoluteString)
                     PaymentNavigationHelper.openPaymentSheet(
@@ -462,7 +476,6 @@ struct MainNavView: View {
             }
     }
 
-    @ViewBuilder
     private func missingPendingImportView(fallbackRoute: Route) -> some View {
         Color.customBlack
             .task {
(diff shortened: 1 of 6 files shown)


piotr-iohk (commented):
Verified success, cancel and error callbacks. All work as expected.

Note the callbacks are available in `0.28` pubky-ring version which is in TestFlight.
Tested also with `0.27` pubky-ring (without callbacks). The flow works as expected as well (obv no callbacks, also no regressions).

piotr-iohk:
This might be un-related to this change. 


Observed few times after a moment of inactivity in Bitkit (with pubky-ring profile imported):

[bitkit_logs_2026-04-29_09-35-29.zip](https://github.com/user-attachments/files/27197205/bitkit_logs_2026-04-29_09-35-29.zip)

Screenshot 2026-04-29 at 11 32 03

After that the profile is disconnected.


https://github.com/user-attachments/assets/91a897f2-b595-4a1d-b3c8-9699f85787c2

jvsena42 (approved):
Tested: approve and deny options. both returned no Bitkit app

piotr-iohk:
Retesting after recent updates.

I think there is some issue with `x-cancel` and `x-error`.

 - on Deny in Pubky-ring we get back to Bitkit - but we see "Waiting for authorization" infinitely, should get back to initial state
 - on `x-error` it is the same ("Waiting for authorization" infinitely) and the error message is not shown - (I believe it was presented in toast, when I tested before)

See the recodring - first is "Deny" then Authorization attempt on broken pubky.


https://github.com/user-attachments/assets/4027857e-dcd4-4229-b2d2-45e83d824706

piotr-iohk (changes requested):

ben-kaufman:
@piotr-iohk can you please attach logs from it?

piotr-iohk:
> @piotr-iohk can you please attach logs from it?

[bitkit_logs_2026-04-30_16-37-01.zip](https://github.com/user-attachments/files/27251103/bitkit_logs_2026-04-30_16-37-01.zip)

ben-kaufman:
Checked the logs. This looks like a Pubky Ring callback encoding issue.

Bitkit sends callback URLs with a nonce value, e.g. `bitkit://pubky-auth/cancel?nonce=<uuid>`, but the app is receiving callbacks like:

- `bitkit://pubky-auth/cancel?nonce`
- `bitkit://pubky-auth/error?nonce&errorCode=AUTH_FAILED&errorMessage=...`

So the `nonce` key is preserved, but its value is dropped. Because custom URL scheme callbacks are unauthenticated on iOS, Bitkit intentionally rejects callbacks unless the nonce value matches the active auth attempt. Pubky Ring should echo the full x-callback URL verbatim, including `nonce=<uuid>`, for `x-success`, `x-cancel`, and `x-error`.

I also fixed Bitkit so malformed cancel/error callbacks no longer leave the UI stuck on “Waiting for authorization.” We still treat those callbacks as untrusted: malformed success is ignored, and malformed error does not display attacker-controlled text.

piotr-iohk:
Created issue on pubky-ring: 
 - https://github.com/pubky/pubky-ring/issues/265

piotr-iohk (approved):
LGTM, my previous issues no longer observed.

piotr-iohk:
@ben-kaufman just seen some issues already fixed here and realized this didn't get into `master` yet.
Just a friendly reminder that we're going to need `codex/pr-527-contact-fixes` 👍

PR #540, opened 2026-05-05
[codex] Prevent dust change from triggering send-all

ben-kaufman:
## Summary

- prevent dust-change detection from turning a normal on-chain send into a send-all transaction
- require max sends to still match the freshly calculated max-send amount before using `sendAllToAddress`
- keep transfer send-all behavior explicit and add a regression test for the reported dust-change values

## Root Cause

The iOS send confirmation flow treated dust-change handling as equivalent to an explicit max send. When selected UTXOs left dust change, `shouldUseSendAll` could become true and the send path called `sendAllToAddress`, draining the wallet instead of sending the requested amount.

## Validation

- `swiftformat Bitkit/Utilities/DustChangeHelper.swift Bitkit/Views/Transfer/SpendingConfirm.swift Bitkit/Views/Wallets/Send/SendConfirmationView.swift BitkitTests/DustChangeHelperTests.swift`
- `git diff --check`
- `xcrun swiftc -parse Bitkit/Utilities/DustChangeHelper.swift BitkitTests/DustChangeHelperTests.swift`

Fixes #536

### DIFF
--- a/Bitkit/Utilities/DustChangeHelper.swift
+++ b/Bitkit/Utilities/DustChangeHelper.swift
@@ -1,15 +1,15 @@
 import Foundation
 
-/// Helper for determining when to use sendAll to avoid creating dust change outputs.
+/// Helper for detecting dust change during on-chain sends.
 enum DustChangeHelper {
-    /// Returns true if the expected change would be dust (below dust limit), so sendAll should be used.
+    /// Returns true if the expected change would be dust (below dust limit).
     /// - Parameters:
     ///   - totalInput: Total sats from selected UTXOs (or spendable balance)
     ///   - amountSats: Amount to send to recipient
     ///   - normalFee: Fee for a normal send (recipient + change outputs)
     ///   - dustLimit: Minimum non-dust amount (default: Env.dustLimit)
-    /// - Returns: true when change would be dust and sendAll should be used
-    static func shouldUseSendAllToAvoidDust(
+    /// - Returns: true when change would be dust
+    static func wouldCreateDustChange(
         totalInput: UInt64,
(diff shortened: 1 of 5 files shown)


piotr-iohk:
Tested with steps to reproduce from the issue: https://github.com/synonymdev/bitkit-ios/issues/536
 - using manual coin selection - sent 5000 onchain
 - selected uxtos 2640 + 2810 =  5450 
 - send -> Insufficient funds toast error message
(not ideal but better than sendAll) 


The same scenario on current `master` sends out all funds.

**Side note:** Android enforces selecting another output in that case - and displays something like Total required which seems to account for potential dust, I assume that auto coin selection follows the same on Android and is a bit better than iOS?

Screenshot 2026-05-05 at 18 02 59

Compare total required with iOS's (master and current), Total required is 5000, so I guess it will produce "Insuficient funds" more frequently:

Screenshot 2026-05-05 at 18 04 41

pwltr:
Tested onchain send near max amount for manual and autopilot coin selection. Both result in an error which I think was the intention for this hotfix.

Manual ✅ 

https://github.com/user-attachments/assets/ceb23d31-f61e-44a1-8b8c-ddfa44638cc7

Autopilot ✅ 

https://github.com/user-attachments/assets/f0e9ce13-e021-4afa-9390-280a0c50f26a

Exact max amount send ✅ 

Transfer to spending still behaves the same, although 'available' amount seems higher than max amount (not sure if intended behavior) ✅

pwltr (approved):

piotr-iohk (approved):

ovitrif (approved):

PR #570, opened 2026-05-26
fix: localize copyright year

piotr-iohk:
### Description

Updates the Support page copyright text so it no longer hardcodes `2025`.

- Adds `settings__support__copyright` to the English localization source with the shared Transifex `{year}` placeholder.
- Uses the existing localization helper to render the copyright line with the current calendar year.
- Leaves non-English localization files untouched so Transifex remains the source of truth for translated values.
- Adds a changelog fragment for the user-visible Support page copy fix.

Transifex note: this PR only changes the English source string. After merge or before release, push source to Transifex and pull translated strings once translations are ready.

### Linked Issues/Tasks

N/A

### Screenshot / Video

TBD

### QA Notes

- Open Settings > Support and verify the copyright line shows the current year.
- Verified `plutil -lint Bitkit/Resources/Localization/*.lproj/Localizable.strings` passes.
- Verified `git diff --check` passes.
- Verified Cursor lints reported no errors.
- `node scripts/validate-translations.js` could not run locally because the checkout is missing the `glob` Node dependency.
- Not run: full Xcode build.


ovitrif (commented):
utAck, see codex review 🙏🏻

ovitrif (approved):
utAck ✅

PR #578, opened 2026-06-04
fix(lightning): copy connection detail IDs

### LINKED ISSUE #355: [BUG] Unable to copy the channel id/point, and peer id in Connection
I can't copy these values. I had a force close and now need to type these values manually :|

CypherPoet:
Fixes #355

### Description

Makes the channel ID, channel point, and node ID copyable on the Lightning connection details screen. These values were display-only and truncated in the middle, so anyone recovering from a force close had no way to retrieve them without retyping by hand.

Tapping any of the three identifier rows now copies the full value, with a haptic tap and a "Copied To Clipboard" toast. This matches the copy behavior already used elsewhere in the app (e.g. the address viewer). Amount, date, and status rows stay non-interactive.

### Linked Issues/Tasks

#355

### Screenshot / Video

https://github.com/user-attachments/assets/6a3c8594-a497-402f-9bdf-e442192338aa

#### Manual Tests
- [x] **1.** Settings → Advanced → Lightning Connections → tap a channel → tap Channel ID: "Copied To Clipboard" toast + haptic; pasted value is the full, untruncated ID.
  - [x] **1b.** tap Channel Point: copies `txid:vout`.
  - [x] **1c.** tap Node ID: copies the full counterparty node ID.
- [x] **2.** `regression:` same screen → tap an amount / date / status row: nothing copies, no toast.
#### Automated Checks
- Built for iOS Simulator (Debug) locally: BUILD SUCCEEDED, the change compiles clean with no new warnings and the app launches.
- CI: standard build/test checks run by the PR bot.


ovitrif (dismissed):
Hi @CypherPoet 👋🏻 

Thank you for your first contribution 🙏🏻 , I checked and found the [android app](https://github.com/synonymdev/bitkit-android) to already allow the same functionality you added in this PR; so this is not only a user-facing improvement but also an improvement on feature parity and consistency across both apps.

---

Review:

tAck 👏🏻 

Would be great if you can add the same functionality when user taps one of the 2 IDs listed at the top of the details rows, namely Order ID and Transaction.

|  |
| - |

On Android tapping Transaction row also opens the block explorer in the browser, but we can tackle that in a separate PR to bring parity on iOS with that behavior.

| Android Recording |
| - |
| <video src="https://github.com/user-attachments/assets/c0e2e1f8-4153-4be9-b03c-239a65b60867"> |

One final nit, it would be great if you could add the same tap pressable effect that Android has, if iOS already has some building blocks ready for this, so the tap feels less rigid on UI, although UX wise the haptics are quite nice already and do cover for that responsiveness goal, to some extent.

PS. Unfortunately we just cut the release branch for next week's release earlier today, so this could only land in the release after that, somewhere early next month. It's unfortunate because if it was opened one day earlier, it would've been live in production next week 🤦🏻 .

ovitrif (approved):
tAck, retested.

Not sure `@ViewBuilder` is still recommended, I remember vaguely many such annotations were removed with latest swiftlint update, but unless I see a lint issue in CI I will treat it as ok.

PR #583, opened 2026-06-05
fix: confirm private paykit toggles

ben-kaufman:
Fixes #582

### Description

This PR fixes private contact payment preferences so the private sharing toggle only reflects confirmed endpoint state. Enabling private sharing now requires immediate publication for saved contacts before the setting is shown as enabled, and disabling private sharing only flips the setting off after endpoint removal succeeds.

If Paykit has to defer or retry the private publish path, the toggle stays on its previous value and shows an error instead of claiming the preference was applied.

### Linked Issues/Tasks

Fixes #582

### Screenshot / Video

N/A

#### Manual Tests
N/A

#### Automated Checks
- `swiftformat Bitkit/Views/Settings/General/PaymentPreferenceView.swift Bitkit/Services/PrivatePaykitService+Contacts.swift`
- `swiftformat --lint Bitkit/Views/Settings/General/PaymentPreferenceView.swift Bitkit/Services/PrivatePaykitService+Contacts.swift`
- `git diff --check`
- `xcodebuild -project Bitkit.xcodeproj -scheme Bitkit -configuration Debug -destination 'platform=iOS Simulator,name=iPhone 17' -only-testing:BitkitTests/PrivatePaykitServiceTests -derivedDataPath /private/tmp/bitkit-ios-582-dd test` passed with 29 tests.


ovitrif (approved):
tAck

PR #585, opened 2026-06-05
fix: cap number pad entry at available balance across amount screens

CypherPoet:
### Description

Caps number-pad entry at the contextual maximum across the amount-entry screens, so you can no longer type an amount you cannot send. Originally split across three PRs, this work was consolidated here per [maintainer request](https://github.com/synonymdev/bitkit-ios/pull/585#issuecomment-4659985810) to keep the E2E updates pointed at a single PR (supersedes #584, folds in #586).

- **Send amount** (#584's commits): introduces `maxAmountOverride` on `AmountInputViewModel` and caps the send screen at the available balance.
- **Remaining screens with a contextual maximum** get the same cap: LNURL pay, LNURL withdraw, transfer-to-spending, advanced spending, and manual external channel funding.
- **Over-cap deletion**: deletions always apply, even when the current amount sits above the cap (for example a prefilled invoice over the available balance), so the amount can be reduced instead of trapping the user.
- **Manual external funding gap**: Continue stayed enabled for amounts above the fundable balance; that screen is now gated on the available balance like the others.
- **Refactor** (folded in from #586): migrates `AmountInputViewModel` from `ObservableObject` to `@Observable`.

The receive and CJIT amount screens are intentionally left uncapped: a receive amount has no balance ceiling.

### Linked Issues/Tasks

Related: #346
Supersedes: #584, #586

### Screenshot / Video

https://github.com/user-attachments/assets/fd56b1ed-6305-487a-bf28-e5fea87f1b85

#### Manual Tests
- [ ] **1.** Send → Amount: type past the available balance: amount freezes at the max.
- [ ] **2.** With an amount above the cap (e.g. prefilled from an invoice over the balance): delete still reduces the amount; typing more digits stays blocked.
- [ ] **3.** Transfer → Spending Amount: type past Available: amount freezes at the max.
- [ ] **4.** Spending Advanced: type past the max LSP balance: capped.
- [ ] **5a.** Fund Manual (external node) → Amount: type past Available: capped.
  - [ ] **5b.** Fund Manual → enter an amount above the fundable balance: Continue is disabled (previously enabled).
- [ ] **6.** LNURL Pay → Amount: type past max sendable: capped.
- [ ] **7.** LNURL Withdraw → Amount: type past max withdrawable: capped.
- [ ] **8.** `regression:` Receive / CJIT Amount: no upper cap; any amount can still be entered.
#### Automated Checks
[...]

PR #592, opened 2026-06-10
refactor: spring-in, fade-out toast motion

CypherPoet:
### Description

This PR aims to establish a more-polished motion design for the app's toast show/hide motion, which previously animated with a flat ease (`easeInOut`, 0.4s) that felt a tad mechanical compared to the spring motion used across the rest of the app (including the toast's own drag-to-dismiss snap-back).

I went with a behavior that's asymmetric, mirroring Apple's system banner pattern:

1. **Arrival is physical**: the toast materializes in place, settling down from 12pt above its resting position with an opacity fade, and a slight 0.98 → 1.0 scale, driven by `.snappy(duration: 0.4)` (a gently damped spring preset). This replaces the full slide-in from offscreen, which swept the whole banner across the top of the screen.
2. **Departure gets out of the way**: a quick fade (`easeOut`, 0.2s). By dismissal time the toast has been read, so exit motion is kept to a minimum.
3. The drag-to-dismiss snap-back now uses the same entrance curve, so the component speaks one motion dialect.

The motion constants live in a single `ToastMotion` namespace. And I wanted to make them a convenient starting point for any feedback from the design team:

| Knob | Value |
| --- | --- |
| Entrance curve | `.snappy(duration: 0.4)` |
| Settle distance (from above) | 12pt |
| Entrance scale | 0.98 → 1.0 |
| Exit curve | `.easeOut(duration: 0.2)` |

**Note on cross-platform consistency:** Android currently slides the full toast in and out from the top edge with a fade, symmetric in both directions (`fadeIn() + slideInVertically` / `fadeOut() + slideOutVertically` in `ToastView.kt`), and its drag snap-back uses a 0.7-damping spring, the same entrance/gesture mismatch this PR addresses on iOS. If this PR goes forward, a counterpart Android PR maps 1:1 (`fadeIn + slideInVertically(-12dp) + scaleIn(0.98)` in, `fadeOut` out, snap-back damping aligned), and I can take that on as a follow-up.

### Linked Issues/Tasks

Follow-up to the motion conversation in #585: https://github.com/synonymdev/bitkit-ios/pull/585#issuecomment-4673702300

### Screenshot / Video

https://github.com/user-attachments/assets/d652e597-e06c-4a44-a32b-0418823d11e1

Feels a bit less like the toast is being rammed into your face, IMO 😂 ... also plays nicely with "standard", informational toast messages:

https://github.com/user-attachments/assets/7d3804e4-e573-4a43-ad14-0a3d657fbfed

#### Manual Tests
- [ ] **1.** Settings → Advanced → Address Viewer → long-press an address: Copied toast settles in from just above, auto-hides after 4s with a quick fade.
- [ ] **2.** `regression:` toast visible → tap content beneath the toast: taps pass through, and the area is tappable again after dismiss.
- [ ] **3a.** `regression:` drag the toast upward: dismisses.
  - [ ] **3b.** small downward drag → release: snaps back and auto-hide resumes.
#### Automated Checks
- No automated coverage exists for toast animation; verified by simulator recording (before/after videos above).

### DIFF
--- a/Bitkit/Components/ToastView.swift
+++ b/Bitkit/Components/ToastView.swift
@@ -21,7 +21,7 @@ struct ToastView: View {
         .frame(maxWidth: .infinity, alignment: .leading)
         .padding(16)
         .background(accentColor.opacity(0.32))
-        .background(.ultraThinMaterial)
+        .background(BlurView())
         .cornerRadius(16)
         .shadow(color: .black.opacity(0.4), radius: 10, x: 0, y: 25)
         .accessibilityIdentifierIfPresent(toast.accessibilityIdentifier)
--- a/Bitkit/Components/ToastView.swift
+++ b/Bitkit/Components/ToastView.swift
@@ -78,7 +78,7 @@ struct ToastView: View {
                         }
                     } else {
                         // Snap back to original position
-                        withAnimation(.spring(response: 0.3, dampingFraction: 0.7)) {
+                        withAnimation(ToastMotion.entrance) {
                             dragOffset = 0
                         }
                     }
--- a/Bitkit/Managers/ToastWindowManager.swift
+++ b/Bitkit/Managers/ToastWindowManager.swift
@@ -5,9 +5,11 @@ import UIKit
 /// gets out of the way (a quick fade), mirroring Apple's system banner pattern.
 enum ToastMotion {
     static let entrance: Animation = .snappy(duration: 0.4)
-    static let exit: Animation = .easeOut(duration: 0.2)
-    /// Hit-test frame cleanup waits for the exit to finish.
-    static let exitSettleTime: Double = 0.3
+    static let exit: Animation = .easeOut(duration: exitDuration)
+    /// Hit-test frame cleanup waits for the exit to finish, with a small buffer.
+    static let exitSettleTime: Double = exitDuration + 0.1
+
+    private static let exitDuration: Double = 0.2
 }
 
 @MainActor
@@ -253,9 +255,10 @@ struct ToastWindowView: View {
                         .allowsHitTesting(false) // Spacer doesn't intercept touches
                 }
                 .id(toast.id)
-                // Materialize in place: a short rise with a fade and a slight scale-up reads as
-                // "arriving" without sweeping the whole banner across the screen. Departure is a
-                // plain fade; the toast has been read by then, so any exit motion is just noise.
+                // Materialize in place: the toast settles down from 12pt above its resting
+                // position with a fade and a slight scale-up, reading as "arriving" without
+                // sweeping the whole banner across the screen. Departure is a plain fade; the
+                // toast has been read by then, so any exit motion is just noise.
                 .transition(
                     .asymmetric(
                         insertion: .offset(y: -12).combined(with: .opacity).combined(with: .scale(scale: 0.98)),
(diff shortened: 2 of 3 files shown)


Bitkit/Managers/ToastWindowManager.swift
```diff
@@ -1,6 +1,15 @@
 import SwiftUI
 import UIKit
 
+/// Motion for toast show/hide. Arrival is physical (a gentle spring settle); departure just
+/// gets out of the way (a quick fade), mirroring Apple's system banner pattern.
+enum ToastMotion {
+    static let entrance: Animation = .snappy(duration: 0.4)
+    static let exit: Animation = .easeOut(duration: 0.2)
+    /// Hit-test frame cleanup waits for the exit to finish.
+    static let exitSettleTime: Double = 0.3
```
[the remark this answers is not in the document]
CypherPoet: Addressed in f27ebfc4 by deriving the value instead of documenting the invariant: the exit animation and `exitSettleTime` are now both built from a single `exitDuration` constant, so they can no longer drift apart.

Bitkit/Managers/ToastWindowManager.swift
```diff
@@ -244,15 +253,25 @@ struct ToastWindowView: View {
                         .allowsHitTesting(false) // Spacer doesn't intercept touches
                 }
                 .id(toast.id)
-                .transition(.move(edge: .top).combined(with: .opacity))
+                // Materialize in place: a short rise with a fade and a slight scale-up reads as
+                // "arriving" without sweeping the whole banner across the screen. Departure is a
+                // plain fade; the toast has been read by then, so any exit motion is just noise.
+                .transition(
+                    .asymmetric(
+                        insertion: .offset(y: -12).combined(with: .opacity).combined(with: .scale(scale: 0.98)),
```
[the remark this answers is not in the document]
CypherPoet: Good catch, the motion is a settle from above, not a rise. Fixed the code comment and the PR description wording in f27ebfc4. The `-12` offset direction itself is intentional: a top-edge banner should arrive from the edge it lives at.

Bitkit/Components/ToastView.swift
pwltr: I noticed the colors don't really align with the design prototype. Could you also change this to use the custom `BlurView` instead?

```suggestion
        .background(BlurView())
```

Image

pwltr (changes requested):

aldertnl (dismissed):
LGTM

Bitkit/Components/ToastView.swift
CypherPoet: > I noticed the colors don't really align with the design prototype. Could you also change this to use the custom `BlurView` instead?
> 
> Image

Applied in e49634c2. The custom BlurView keeps only the gaussian blur, so the accent overlay is no longer washed out by the material's built-in tint and saturation. Verified on the simulator against the prototype: success and warning toasts now match it exactly (#75BF72 and #FF4400 titles).


pr592-after-error-toast

pr592-after-success-toast


One follow-up question on color alignment: The app's .error toast type uses redAccent (#E95164), a crimson that doesn't appear anywhere in the prototype; the prototype's failure toasts ("Spending Balance Setup Failed", "Instant Payments Unavailable") all use the brand orange (#FF4400). Should error toasts switch to the brand color too, or does the design keep a separate red for errors? I held off on changing that because I wasn't sure.

pwltr: Asked design and tracked separately [here](https://github.com/synonymdev/bitkit-ios/issues/593)

pwltr (approved):

PR #595, opened 2026-06-12
fix: cap Transfer to Spending max at LSP limit

jvsena42:
This PR brings the Transfer to Spending amount screen to parity with bitkit-android #908:

1. Fixes the transferable maximum collapsing to zero when the on-chain balance exceeds the LSP's max channel size.
2. Aligns the displayed "Available" balance with the transferable maximum.
3. Disables the number pad while the maximum is still being calculated.
4. Calculates the 25% button over the final after-fee maximum.

### Description

On Transfer to Spending, an on-chain balance larger than the LSP's max channel size made the liquidity calculation report a zero max client balance (the balance already saturates the channel), so the spendable maximum collapsed to zero and the user was told they could transfer nothing while their funds stayed on-chain. The prospective client balance is now clamped to the LSP's max client balance before liquidity options and the order fee are estimated, so the maximum caps at that limit instead of dropping to zero.

On top of that fix, three smaller alignments with the Android screen:

- The "Available" row now shows the after-fee transferable maximum, so Available, the MAX button, and the input cap all agree on the same value.
- The number pad is dimmed and ignores input while the maximum is being (re)calculated, and re-enables once the limit is ready.

### Linked Issues/Tasks

[Figma](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=40364-128036&m=dev) 

#### Manual Tests
- [x] **1.** Fund savings above the LSP max channel size → Transfer → Transfer to Spending: max and Available show the LSP client-balance limit (not ₿ 0); Continue is enabled.
- [x] **2.** Transfer → Transfer to Spending: while the max loads, the number pad is dimmed and ignores taps, then enables once Available appears.
- [x] **3.** Transfer to Spending → tap MAX: fills the after-fee max; Available matches it.
- [x] **4.** `regression:` normal (below-cap) balance → Transfer to Spending: max / Available / MAX still equal the after-fee value; 25% and MAX work.

#### Automated Checks
- Unit tests added: cover the LSP client-balance clamp in `BitkitTests/TransferViewModelTests.swift` (caps at LSP max when on-chain exceeds it, uses the full balance when LSP info is unavailable, and returns zero when liquidity reports zero).
- Local: `xcodebuild test -only-testing:BitkitTests/TransferViewModelTests` passed and the app target builds clean.

### Screenshot / Video

https://github.com/user-attachments/assets/2641c96f-eb10-4425-9044-d2cbc3ef890e

https://github.com/user-attachments/assets/85305b48-11be-4922-9219-55d6ca0dd451


Bitkit/ViewModels/TransferViewModel.swift
```diff
@@ -360,6 +360,45 @@ class TransferViewModel: ObservableObject {
...
+        lspMaxClientBalance: UInt64?,
+        transferValues: (_ clientBalance: UInt64) -> TransferValues,
+        estimateOrderFee: (_ clientBalance: UInt64, _ lspBalance: UInt64) async throws -> (networkFeeSat: UInt64, serviceFeeSat: UInt64)
+    ) async rethrows -> (available: UInt64, max: UInt64) {
+        // First pass: estimate the LSP fee against the full on-chain balance.
+        let values1 = transferValues(onchainAvailable)
+        let lspBalance1 = max(values1.defaultLspBalance, values1.minLspBalance)
+        let fee1 = try await estimateOrderFee(onchainAvailable, lspBalance1)
+        let initialFees = fee1.networkFeeSat + fee1.serviceFeeSat
+        let balanceAfterLspFee = onchainAvailable > initialFees ? onchainAvailable - initialFees : 0
```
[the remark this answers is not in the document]
jvsena42: that is ok. I would happen when onchainAvailable is only slightly above the cap. the result would be an value slightly bellow the max, still conservative

Bitkit/Views/Transfer/SpendingAmount.swift
```diff
@@ -120,8 +146,7 @@ struct SpendingAmount: View {
 
             NumberPadActionButton(text: t("lightning__spending_amount__quarter")) {
                 guard let max = maxTransferAmount else { return }
-                let quarter = UInt64(wallet.spendableOnchainBalanceSats) / 4
-                amountViewModel.updateFromSats(min(quarter, max), currency: currency)
+                amountViewModel.updateFromSats(max / 4, currency: currency)
```
pwltr: This should be 25% of savings balance, as it is currently

pwltr (changes requested):

Bitkit/Views/Transfer/SpendingAmount.swift
jvsena42: https://github.com/user-attachments/assets/6c4b7a60-d9bf-40c1-afc2-effa3d1c0c99

Behaviour restored. I'll match android too
Still seems strange having the same value for the quarter and the max for some cases

jvsena42: Left a note in Figma https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ?node-id=40364-128036&m=dev#1810552156

pwltr (approved):

PR #597, opened 2026-06-13
fix: restore Explore button on transfers

### LINKED ISSUE #361: [Bug] Force Close TX showing in the LN activity template
I got a force close. The close shows in the activity but it is shown with the LN transaction template.
In reality, this is a onchain transaction though so it would be nice to also see the transaction id and so on.

CypherPoet:
Fixes #361

This PR restores access to the on-chain details for transfer activities, including force closes.

### Description

A force close, like any channel close, settles on-chain, so Bitkit already stores it as an on-chain activity with a transaction id, inputs, outputs, and a block explorer link. Those details had become unreachable from the activity screen: the transfer detail layout replaced the "Explore" button with a "Connection" button, and "Connection" only opens the channel, not the transaction. Issue #361 shows a force close where there is no way to see the transaction id.

This adds a dedicated full-width "Explore" row for transfer activities, shown alongside the existing "Connection" button, so the on-chain transaction details are reachable again. Regular on-chain and Lightning activities are unchanged. The behavior applies to every transfer because a stored on-chain activity does not record whether a close was cooperative or forced, and both are real on-chain transactions with a transaction id worth surfacing.

### Linked Issues/Tasks

#361 - [Bug] Force Close TX showing in the LN activity template

### Screenshot / Video

Before (issue #361) on the left, after on the right:

issue361-before-after

#### Manual Tests
- [ ] **1.** open a connection → close it → Activity → tap the From Spending transfer → Activity Item: shows Boost, Connection, and a new full-width Explore button.
- [ ] **2.** Activity Item (transfer) → Explore → Activity Explorer: shows transaction id, inputs, outputs, and Open Block Explorer.
- [ ] **3.** Activity Item (transfer) → Connection: still opens the connection detail.
- [ ] **4.** `regression:` received on-chain activity → Activity Item: Explore appears in its usual slot, with no extra row.
- [ ] **5.** `regression:` LN payment → Activity Item: unchanged; Explore opens preimage and invoice details.
#### Automated Checks
- BitkitTests unit suite passes locally (activity storage and the other unit suites, 0 failures). Two `AddressTypeIntegrationTests` fail only because the live Blocktank regtest deposit endpoint returns 404; they depend on that external service and are unrelated to this change.
- Verified the full flow live on the iOS simulator against regtest: opened a connection, closed it, and confirmed the From Spending activity now reaches the transaction details through Explore.

### DIFF
--- a/Bitkit/Views/Wallets/Activity/ActivityItemView.swift
+++ b/Bitkit/Views/Wallets/Activity/ActivityItemView.swift
@@ -566,26 +566,22 @@ struct ActivityItemView: View {
                 }
                 .accessibilityIdentifier(boostButtonIdentifier)
 
-                if isTransfer, let channelId = transferChannelId {
-                    CustomButton(
-                        title: t("lightning__connection"), size: .small,
-                        icon: Image("bolt-hollow")
-                            .foregroundColor(accentColor),
-                        shouldExpand: true
-                    ) {
-                        navigation.navigate(.connectionDetail(channelId: channelId))
-                    }
-                    .accessibilityIdentifier("ChannelButton")
-                } else {
-                    exploreButton
-                }
+                exploreButton
             }
             .frame(maxWidth: .infinity)
 
-            // Transfers replace the Explore button above with Connection, so give them their own
-            // Explore row to reach the on-chain details (tx id, block explorer) of the channel transaction.
-            if isTransfer, transferChannelId != nil {
-                exploreButton
+            // Transfers get a dedicated full-width Connection row to their channel, keeping the Explore
+            // button in the same slot as every other on-chain activity (tx id, block explorer) for consistency.
+            if isTransfer, let channelId = transferChannelId {
+                CustomButton(
+                    title: t("lightning__connection"), size: .small,
+                    icon: Image("bolt-hollow")
+                        .foregroundColor(accentColor),
+                    shouldExpand: true
+                ) {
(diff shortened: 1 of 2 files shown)


jvsena42 (changes requested):
https://github.com/synonymdev/bitkit-ios/pull/597#issuecomment-4720605018

Bitkit/Views/Wallets/Activity/ActivityItemView.swift
```diff
@@ -566,33 +566,39 @@ struct ActivityItemView: View {
...
-                        navigation.navigate(.activityExplorer(viewModel.activity))
-                    }
-                    .accessibilityIdentifier("ActivityTxDetails")
-                }
+                exploreButton
             }
             .frame(maxWidth: .infinity)
+
+            // Transfers get a dedicated full-width Connection row to their channel, keeping the Explore
+            // button in the same slot as every other on-chain activity (tx id, block explorer) for consistency.
```
jvsena42: This comment could be removed as per #602

jvsena42 (changes requested):
Code is good, just need one change for avoid pollution

Bitkit/Views/Wallets/Activity/ActivityItemView.swift
CypherPoet: Removed in 1079e788 👍

jvsena42 (approved):
tAck

PR #598, opened 2026-06-15
fix: reset suggestions widget when re-added empty

jvsena42:
Fixes #596

This PR fixes the Suggestions widget disappearing after every card is dismissed, both when the widget is re-added and while editing the home grid.

### Description

When every suggestion card is dismissed with the [x] button, the Suggestions widget removes itself from the home grid, which is expected. Two follow-on problems remained, both caused by the dismissed cards being persisted so the widget kept resolving to an empty set of cards:

1. Re-adding the widget (after deleting it in edit mode) never brought it back visually — it only appeared while editing. Re-adding now clears the dismissed cards when the widget would otherwise show nothing, restoring the default set of suggestions. This reuses the same reset as the "Reset Suggestions" action in Settings, and only runs when the suggestions are empty, so re-adding while some cards are still visible leaves those dismissals untouched.

2. With all cards dismissed, the widget was kept in the grid while editing but rendered nothing, leaving an invisible row that could not be reordered or removed. When editing an empty Suggestions widget it now falls back to the static preview set (the same cards shown in the widget preview), so it stays visible and removable. This fallback only happens in edit mode when the suggestions are empty; live cards and other widget types are unaffected.

### Linked Issues/Tasks

Fixes #596 - Suggestion widget bug when empty and re-adding

### Screenshot / Video

https://github.com/user-attachments/assets/3a95638b-4877-4195-903d-c4d08315048e

https://github.com/user-attachments/assets/d1f348c5-757d-48ee-bb03-b9d927b79591

#### Manual Tests
- [x] **1.** Home → dismiss every Suggestions card with [x]: widget hides once the last card is dismissed.
- [x] **2.** edit mode → remove the Suggestions widget → add it back via Widgets list → preview → Save Widget: widget reappears showing the default suggestion cards.
- [x] **3.** dismiss every card (do not remove widget) → enter edit mode: widget shows the static preview set and can be reordered/removed, instead of an invisible row.
- [x] **4a.** `regression:` dismiss only some cards → remove and re-add the Suggestions widget: remaining cards stay as-is, dismissed ones are not restored.
  - [x] **4b.** `regression:` some cards still visible → enter edit mode: live cards show as normal (no preview fallback).

#### Automated Checks
N/A (no automated coverage changed; verified by code inspection and SwiftFormat)

### DIFF
--- a/Bitkit/Components/Widgets/Suggestions.swift
+++ b/Bitkit/Components/Widgets/Suggestions.swift
@@ -164,6 +164,10 @@ struct Suggestions: View {
     /// When true, show a fixed set of static cards and ignore taps (e.g. widget preview).
     var isPreview: Bool = false
 
+    /// When editing the home grid, keep the widget visible (and reorderable/removable) by falling
+    /// back to the static preview set when there are no live cards to show.
+    var isEditing: Bool = false
+
     var previewCardIds: [String]?
 
     static let previewSheetCardIds = ["backupSeedPhrase", "pin", "transferToSpending", "support"]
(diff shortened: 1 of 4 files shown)


piotr-iohk (approved):
tACK

Issue #589, opened 2026-06-08
epic: Trezor Hardware Wallet UI


ovitrif:
**Design:** [Figma — Trezor Screens](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=44997-151646)

**Related (Android):** synonymdev/bitkit-android#998

## Context

Bitkit adds a **Hardware Wallet** balance type alongside Savings (on-chain) and Spending (Lightning), backed by Trezor. The hardware balance is a watch-only view of funds held on a paired Trezor: Bitkit displays the amount and the device's on-chain activity and never holds the keys. Removing a device stops Bitkit from displaying that balance.

Paired Trezor identities appear on Home, Hardware Wallet detail, Settings, and in the Connect Hardware flow. Transfer to Spending from a hardware wallet is a send signed on the device.

Trezor v2 (milestone 2.5.0) extends that surface:

- Send on-chain from a paired Trezor through the normal Send flow, signing on the device.
- Receive on-chain to a paired Trezor, including on-device address verification.
- Pair passphrase (hidden) wallets as separate watch-only identities.
- Keep hardware activity, tags, and custom names across backup and restore.

### Device naming & connection indicator

- **Display name** = `label ?: model ?: "Trezor"` — `label` is the device name set on the Trezor (editable on-device / in Trezor Suite), `model` is the hardware model (e.g. "Safe 3" / "Safe 7").
- The Figma **"LABEL FUNDS"** field is a Bitkit-side display name for the balance, defaulting to the resolved device name; persisted per device so the user can rename within Bitkit without touching the device.
- **Connection indicator** = connection type (Bluetooth **or** USB), tinted green while the device is currently connected. Figma examples: [bluetooth](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=44997-151650) · [usb](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=44997-151655).

PR #605, opened 2026-06-22
feat: home-screen hardware wallet (watch-only)

jvsena42:
This PR:
1. Adds watch-only hardware wallet (Trezor) support to the home screen, ported from bitkit-android #999 — Bluetooth only.
2. Bumps bitkit-core 0.1.66 → 0.3.4 and migrates the app to its wallet-scoped activity storage.
3. Refactors the Trezor service layer: extracts a `TrezorManager` from the `TrezorViewModel` god object, adds a fully decoupled `HwWalletManager`, and splits the device-less on-chain calls out of `TrezorService` into a vendor-neutral `OnChainHwService`.

### Description

Brings the first slice of the Trezor hardware-wallet epic to the wallet home screen. A paired Trezor now shows as a watch-only balance: a device row under the Savings/Spending tiles, its sats folded into the headline total, and its on-chain transactions merged into the activity lists with a blue icon. Users with no paired device see a new Hardware suggestion card that opens an intro sheet, and a one-time pairing code is handled by an app-wide Pair Device sheet.

How it works: on connect, each device's account xpubs are captured and persisted. The watch-only layer then runs one on-chain xpub watcher per device and enabled address type, aggregates the per-device balance in memory, and persists each device's activity into bitkit-core scoped by a derived wallet id — so balances and activity stay available while the device is disconnected, and the activity lists merge hardware transactions through the normal pipeline. The same physical device re-paired is de-duplicated by its xpubs into a single tile.

Scope notes:
- Bluetooth only: USB-specific transport, icons, and reconnect handling from the Android PR are intentionally omitted.
- The intro sheet's Continue is intentionally disabled — the connect flow ships in a follow-up; this slice covers the home surface, suggestion card, and the app-wide pairing sheet.
- Adopting wallet-scoped activities required bumping bitkit-core to 0.3.4, which adds a `wallet_id` to every activity. This migrates the whole app (not just hardware) to pass a wallet id through the activity APIs; the normal wallet uses the core default.
- The per-device wallet-id is derived via bitkit-core's `deriveWalletId` — the canonical cross-platform derivation finalized in core 0.3.4 — so iOS and Android produce identical ids for the same device.
- Architecture: the device/connection orchestration + pairing dialogs were extracted out of `TrezorViewModel` into a `TrezorManager`, and the watch-only `HwWalletManager` is driven purely by the composition root (no manager-to-manager or service-to-viewmodel coupling). The device-less on-chain surface — Electrum account/address info, tx history/detail, compose, broadcast, and the xpub watcher — was also split out of `TrezorService` into a vendor-neutral `OnChainHwService` behind an `OnChainWatcherServicing` protocol. `HwWalletManager` now depends on no Trezor-named type; `TrezorService` is reduced to Trezor device FFI (connect, sign, on-device address/pubkey, credentials).

### Linked Issues/Tasks

Ports synonymdev/bitkit-android#999.
Related to #589 

#### Trezor Emulator Setup
Test against the Trezor emulator from [bitkit-docker](https://github.com/ovitrif/bitkit-docker):
1. Start the backend + emulator: clone with `--recurse-submodules`, then `docker compose up -d` and `./scripts/trezor-emulator start`.
2. Run a Debug build (regtest) with these env vars set in the Xcode scheme (Edit Scheme → Run → Arguments → Environment Variables):
   - `TREZOR_BRIDGE=true`
   - `TREZOR_BRIDGE_URL=http://127.0.0.1:21325`
   - Do **not** set `TEST_TREZOR_EMU` — that flag boots the standalone dev/test harness and bypasses the wallet.
3. Enable Dev Settings, then open Settings → Advanced → Trezor Hardware Wallet.

#### Manual Tests
- [x] **1.** No paired device → Home: no hardware row, Hardware suggestion card shown.
- [x] **2.** Tap Hardware suggestion card → Hardware intro in-sheet opens: device visuals shown, Cancel works, Continue disabled.
- [ ] **3.** Connect a Trezor (dev Trezor screen or Bridge emulator) → Home: hardware row appears with device name, blue BTC icon and balance; headline total includes the hardware balance; recent device transactions appear in the activity list with blue icons.
- [ ] **4a.** BLE-connected device → disable phone Bluetooth: indicator turns grey, balance stays visible.
  - [ ] **4b.** re-enable Bluetooth: indicator turns green again on its own (auto-reconnect, no pairing prompt).
- [ ] **5.** Device asks for its one-time pairing code → Pair Device in-sheet opens with the app numpad; entering the code on-screen completes the connect; swipe-down dismiss cancels it.
- [ ] **6.** Tap the hardware row: not-yet-implemented toast shows.
- [ ] **7.** Tap a hardware activity item in Home → Activity Detail opens with a blue icon.
- [ ] **8.** Watcher running → send sats to a Trezor address from an external wallet: received sheet pops up once picked up.
- [ ] **9.** Re-pair the same Trezor → Home shows one tile, balance and activity counted once.
- [x] **10a.** Show All → All Activity: hardware activities listed with blue icons; Sent/Received tabs filter them.
  - [x] **10b.** tag filter: hides hardware activities (they carry no tags).
- [x] **11.** `regression:` open Receive and other large sheets, and the normal on-chain/LN activity list and detail behave as before (wallet-scoped activity migration).
#### Automated Checks
- Unit tests added: `BitkitTests/HwWalletManagerTests.swift` (aggregation, xpub dedup, displayName, merge-by-txid, received-tx detection, monitored-type filtering, watcher params, saturating total, stale-watcher, removeDevice, resetState), `BitkitTests/HwWalletIdTests.swift`, `BitkitTests/HwAddressTypeTests.swift`, `BitkitTests/ActivityHardwareTests.swift`.
- Unit tests modified: `BitkitTests/TrezorViewModelWatcherTests.swift` retargeted to `TrezorManager` after the extraction; the `MockWatcherService` mocks in `TrezorViewModelWatcherTests` and `HwWalletManagerTests` now conform to the new `OnChainWatcherServicing` protocol.
- Ran locally: full `BitkitTests` suite — 553/556 pass; the 3 failures are pre-existing `AddressTypeIntegrationTests` cases that hit the live Blocktank regtest deposit API (HTTP 404), unrelated to this change. `swiftformat --lint`, `scripts/validate-translations.js`, and a Debug build (iPhone 16 simulator) are clean.

### Screenshot / Video

https://github.com/user-attachments/assets/b37868cb-6c30-41d2-a26c-fc34832c7318


Bitkit/Models/HwWalletId.swift
```diff
@@ -0,0 +1,24 @@
...
+import Foundation
+
+/// Derives a stable, app-level wallet id for a paired hardware wallet, used to scope its
+/// activities in bitkit-core's wallet-scoped storage (core 0.3.x).
+///
+/// ⚠️ INTERIM / PROVISIONAL. bitkit-core does not derive this — it only ships the default
+/// `"bitkit"` id (`getDefaultWalletId()`), and the Android app has not yet adopted
+/// wallet-scoped hardware storage. The exact format MUST be agreed cross-platform before
+/// release so iOS and Android produce identical ids for the same device. This is the single
+/// source of the derivation: swap it here when the shared scheme is finalized.
```
jvsena42: attention point

Bitkit/Services/MigrationsService.swift
```diff
@@ -1400,6 +1400,7 @@ extension MigrationsService {
             let invoice = (item.address?.isEmpty == false) ? item.address! : "migrated:\(item.id)"
 
             let lightning = BitkitCore.LightningActivity(
+                walletId: WalletScope.default,
```
jvsena42, line 1403: @ovitrif @coreyphillips What do you think about setting the default Id in the constructor from Bitkit-core?

ovitrif: @jvsena42 I wanted to ask my AI employees ^^ a few days ago when reviewing the actual bitkit-core PR that added this, but then I got sidetracked with other work.

@claude or @codex: is it possible to set defaults params values in Rust and expose that over the FFI layer via UniFFi.

---
`!ai:ignore:` PS. For some reasons these integrations seem to no longer work.

Bitkit/Models/HwWallet.swift
jvsena42, line 1: Maybe HwWallet naming is a bit redundant (Hardware Wallet Wallet)

jvsena42: Maybe not, because are the wallets from the Hardware wallet device

ovitrif: > Maybe not, because are the wallets from the Hardware wallet device
Indeed, HWW is HardWare Wallet, HW is usually HardWare in general terminology 

So HWW is the acronym for the whole thing, at least that's what I infer from discussions with Ben

Bitkit/Services/MigrationsService.swift
ovitrif: from local session:

> yes, UniFFI can expose default values to generated Swift/Kotlin bindings. bitkit-core is using UniFFI 0.29.4 in [Cargo.toml (line 12)](/Users/ovitrif/r/synonym/bitkit-core/Cargo.toml:12), and the current docs support defaults on function args, constructors, records, and enum fields: [UniFFI defaults](https://mozilla.github.io/uniffi-rs/latest/types/defaults.html), [constructors/functions](https://mozilla.github.io/uniffi-rs/latest/proc_macro/functions.html), [records](https://mozilla.github.io/uniffi-rs/latest/proc_macro/records.html).
For LightningActivity.wallet_id, core could add something like #[uniffi(default = "bitkit")] to the record field. The existing #[serde(default = "default_wallet_id")] only helps serde, not the generated Swift initializer. Caveat: Rust still has no native default params; UniFFI generates binding-level defaults and passes the filled value into Rust. If we want one source of truth without duplicating "bitkit" in the attribute, use a core helper/factory that accepts Option<String> and resolves None to DEFAULT_WALLET_ID.

Bitkit/Models/HwWalletId.swift
ovitrif: core walletId derivation APIs exposed over FFI bridge:
- https://github.com/synonymdev/bitkit-core/pull/108

Release:
https://github.com/synonymdev/bitkit-core/releases/tag/v0.3.4

Bitkit/Services/OnChainHwService.swift
```diff
@@ -0,0 +1,200 @@
...
+
+    /// Stop all active watchers.
+    func stopAllWatchers() {
+        onchainStopAllWatchers()
+    }
+}
+
+// MARK: - Network / Electrum helpers
+
+/// Network- and Electrum-derivation helpers shared by all on-chain consumers (`TrezorManager`,
```
ovitrif, line 150: I think this dashboard cleanup now stops every on-chain watcher, not just the dev watcher owned by `TrezorViewModel`. `HwWalletManager` uses the same `OnChainHwService.shared`, so leaving the Trezor dashboard can call `onchainStopAllWatchers()` and leave Home thinking its hardware watchers are still active. Could we make this cleanup stop only the dev watcher id, and reserve the global stop for app reset/wipe?

Bitkit/Managers/HwWalletManager.swift
```diff
@@ -0,0 +1,450 @@
...
+        )
+        let listener = TrezorEventListener { [weak self] id, event in
+            self?.handleWatcherEvent(watcherId: id, event: event)
+        }
+        listeners[spec.watcherId] = listener
+        pendingWatcherStarts.insert(spec.watcherId)
+
+        Task { @MainActor in
+            do {
+                try await watcherService.startWatcher(params: params, listener: listener)
```
ovitrif, line 273: Could we re-check that `spec.watcherId` is still desired after `startWatcher` returns? Right now a device can be forgotten or an address type can be disabled while the start is pending, but this task still inserts the watcher into `activeWatchers`; later events can repopulate a forgotten wallet or stale address type.

Bitkit/AppScene.swift
```diff
@@ -106,6 +114,9 @@ struct AppScene: View {
             .onChange(of: wallet.nodeLifecycleState) { _, newValue in handleNodeLifecycleChange(newValue) }
             .onChange(of: scenePhase, initial: true) { _, newValue in handleScenePhaseChange(newValue) }
             .onChange(of: network.isConnected) { _, isConnected in handleNetworkChange(isConnected) }
+            // Bridge Trezor device state into the watch-only manager without coupling the two:
+            // TrezorManager bumps devicesRevision on any device/connection change.
+            .onChange(of: trezorManager.devicesRevision) { _, _ in pushHardwareDevices() }
```
ovitrif, line 119: This only reconciles hardware watchers when the Trezor device snapshot changes. `HwWalletManager.syncWatchers()` also depends on the current Electrum URL and `addressTypesToMonitor`, so changing either setting can leave existing hardware watchers on the old server or watching removed address types until a device event happens. Could we trigger a hardware watcher sync from those setting changes too?

Bitkit/Managers/HwWalletManager.swift
```diff
@@ -0,0 +1,450 @@
...
+        let sent = transactions.reduce(UInt64(0)) { saturatingAdd($0, $1.sent) }
+        let fee = transactions.compactMap(\.fee).max() ?? 0
+
+        // Classify by core's TxDirection (core 0.3.x): received-vs-sent arithmetic alone cannot
+        // tell a self-transfer/consolidation apart from a normal send (both have sent > received,
+        // since sent == received + fee), so without this a consolidation would show value 0
+        // instead of the fee paid.
+        let txType: PaymentType
+        let value: UInt64
+        switch first.direction {
```
ovitrif: This merged activity takes `direction` from `transactions[0]`, but the same `txid` can appear from multiple watched address types with different wallet-perspective directions. That makes mixed rows depend on dictionary order. Could we derive the final `txType` from the merged `received`/`sent`/`fee` values, or add a mixed-direction test so this stays deterministic?

ovitrif (commented):
Left a few inline notes from the review.

Bitkit/Components/Widgets/Suggestions.swift
```diff
@@ -248,6 +251,7 @@ struct Suggestions: View {
             settings: settings,
             suggestionsManager: suggestionsManager,
             pubkyProfile: pubkyProfile,
+            hasHardwareWallet: isPreview ? false : !hwWalletManager.wallets.isEmpty,
```
ovitrif, line 254: The live `Suggestions` view passes the hardware-wallet state here, but `HomeWidgetsView.widgetsToShow` and `WidgetPreviewSheetView.suggestionsAreEmpty` call `visibleCards` without it. After a hardware wallet is paired, those filters can think the Suggestions widget still has a hardware card while the rendered widget hides it. Could we pass the same `hasHardwareWallet` state into those callers too?

Bitkit/Managers/HwWalletManager.swift
```diff
@@ -0,0 +1,450 @@
...
+                value: value,
+                fee: fee,
+                feeRate: 1,
+                address: "",
+                confirmed: confirmed,
+                timestamp: timestamp,
+                isBoosted: false,
+                boostTxIds: [],
+                isTransfer: false,
+                doesExist: true,
```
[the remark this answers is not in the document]
jvsena42: fixed in https://github.com/synonymdev/bitkit-ios/pull/605/commits/32d491504f2d40bb38698d8f7db3753fb7c91e95

Bitkit/Services/OnChainHwService.swift
jvsena42: fixed https://github.com/synonymdev/bitkit-ios/pull/605/commits/a3a67c6154067be68fb4696247565b79a2e8a671

Bitkit/Managers/HwWalletManager.swift
jvsena42: fixed https://github.com/synonymdev/bitkit-ios/pull/605/commits/8685359e2e71dcdc4ff98f63c093294017e6db5f

Bitkit/AppScene.swift
jvsena42: fixed https://github.com/synonymdev/bitkit-ios/pull/605/commits/d2aa3f87c0b305a4dc9abdb55cd66edf48670c4c

Bitkit/Utilities/AppReset.swift
```diff
@@ -23,6 +23,8 @@ enum AppReset {
         await VssBackupClient.shared.reset()
         VssStoreIdProvider.shared.clearCache()
 
+        OnChainHwService.shared.stopAllWatchers()
```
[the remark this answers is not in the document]
jvsena42: resetState() was unused production code, so I removed it. A clean manager after wipe is already guaranteed structurally: ContentView hosts AppScene().id(session.id) and AppReset.wipe calls session.bump(), which fully recreates AppScene and its @State hwWalletManager with empty state, updateDevices then re-derives live watchers from scratch. So there's no stale activeWatchers/knownDevices to clear. AppReset also still calls OnChainHwService.shared.stopAllWatchers() to tear down the FFI-level loops.

Bitkit/Managers/HwWalletManager.swift
```diff
@@ -0,0 +1,450 @@
...
+        listeners.removeAll()
+        watcherData.removeAll()
+        assignedTimestamps.removeAll()
+        recomputeDerivedState()
+    }
+
+    /// Stop watching a paired hardware wallet and delete its stored activities. The caller is
+    /// responsible for forgetting the device entries (via `TrezorManager.forgetDevice`); the next
+    /// `updateDevices(...)` push then drops it from the tile list.
+    func removeDevice(id deviceId: String) {
```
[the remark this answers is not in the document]
jvsena42: The reset already happens. ContentView hosts the app as AppScene().id(session.id), and AppReset.wipe() calls session.bump() (which assigns a new UUID to session.id). Changing that .id makes SwiftUI tear down and recreate the entire AppScene, including its @State private var hwWalletManager

Bitkit/Components/Widgets/Suggestions.swift
```diff
@@ -34,8 +34,8 @@ enum WalletSuggestionState {
 /// Max 4 cards are shown; when one is dismissed or completed, the next in this list is shown.
 private let suggestionOrderByState: [WalletSuggestionState: [String]] = [
     .empty: ["buyBitcoin", "transferToSpending", "support", "backupSeedPhrase", "pin", "profile", "invite"],
```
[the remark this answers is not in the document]
jvsena42: it is intended to be like this per design https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=40364-133650&m=dev

Bitkit/Managers/HwWalletManager.swift
```diff
@@ -0,0 +1,426 @@
...
+            if isActive, activeWatcherElectrumUrls[spec.watcherId] == spec.electrumUrl { continue }
+            if isActive, !stopActiveWatcher(spec.watcherId) { continue }
+            startWatcher(spec)
+        }
+
+        // A failed stop stays active so the next sync retries it; dropping it here would leave the
+        // orphaned watcher feeding watcherData as a ghost balance.
+        for staleId in activeWatchers.subtracting(desiredIds) {
+            _ = stopActiveWatcher(staleId)
+        }
```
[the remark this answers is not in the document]
jvsena42: Fixed https://github.com/synonymdev/bitkit-ios/pull/605/commits/5777dbbb4e2ca76b925619417caeb3d36cd6877e

Bitkit/Managers/HwWalletManager.swift
```diff
@@ -0,0 +1,494 @@
...
+    func syncWatchers() {
+        let specs = desiredWatcherSpecs()
+        let desiredIds = Set(specs.map(\.watcherId))
+
+        for spec in specs {
+            // A start is already in flight for this watcher; skip so we don't launch a duplicate.
+            // The next sync after it completes reconciles any electrum-url change.
+            if pendingWatcherStarts.contains(spec.watcherId) { continue }
+            let isActive = activeWatchers.contains(spec.watcherId)
+            if isActive, activeWatcherElectrumUrls[spec.watcherId] == spec.electrumUrl { continue }
```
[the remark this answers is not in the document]
jvsena42: https://github.com/synonymdev/bitkit-ios/pull/605/commits/f0d63f8adaacf2c3e52eaf4d5bc22bbfdc632b65

Bitkit/Managers/HwWalletManager.swift
```diff
@@ -0,0 +1,500 @@
...
+        pendingWatcherStarts.insert(spec.watcherId)
+
+        Task { @MainActor in
+            do {
+                try await watcherService.startWatcher(params: params, listener: listener)
+                pendingWatcherStarts.remove(spec.watcherId)
+                // The device may have been forgotten or its address type disabled while the start
+                // was in flight; if this watcher is no longer desired, tear it down instead of
+                // resurrecting a forgotten wallet or a stale address type.
+                guard desiredWatcherIds().contains(spec.watcherId) else {
```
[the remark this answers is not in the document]
jvsena42: fixed https://github.com/synonymdev/bitkit-ios/pull/605/commits/e83dd03d208ede828f61e2729405884e11734e8c

Bitkit/Managers/TrezorManager.swift
```diff
@@ -0,0 +1,673 @@
...
+        // THP reads it via currentSelection() to bind the passphrase at session creation;
+        // non-THP devices re-request it mid-operation and are answered from the same value.
+        uiHandler.setWalletMode(mode, hostPassphrase: passphrase)
+        walletMode = mode
+
+        do {
+            let features = try await trezorService.connect(deviceId: device.path, selection: uiHandler.currentSelection())
+            connectedDevice = device
+            deviceFeatures = features
+            showConfirmOnDevice = false
```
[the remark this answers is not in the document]
jvsena42: Fixed https://github.com/synonymdev/bitkit-ios/pull/605/commits/a80db295b6f5dc219a55c0a13738925dd0eda887

Bitkit/AppScene.swift
```diff
@@ -601,6 +621,9 @@ struct AppScene: View {
         }
 
         if newPhase == .active {
+            // Reconnect a known hardware device so its connection indicator turns green again;
+            // watch-only balances stay live regardless.
+            Task { await trezorManager.autoReconnect() }
```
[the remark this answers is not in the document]
jvsena42: Fixed https://github.com/synonymdev/bitkit-ios/pull/605/commits/eea9bb1ac10f1df5f31f78d83c573bbd9a427f70

piotr-iohk (approved):
Approving.
Remaining items added to: 
 - https://github.com/synonymdev/bitkit-ios/issues/613

PR #619, opened 2026-07-08
fix: migrate legacy backups missing wallet id on restore

coreyphillips:
## Summary

Stacked on #617. Legacy backups that predate wallet-scoped activity data fail to decode on restore because their activity, tag and pre-activity-metadata records have no wallet id, and with bitkit-core 0.4.0 those records require one. The decode throws before the records reach Core.

## Approach

Before decoding a restored envelope, each Core-owned array (`activities`, `activityTags`, `tagMetadata`) is handed to the matching bitkit-core migration helper as raw JSON:

- `migrateBackupActivitiesJson`
- `migrateBackupActivityTagsJson`
- `migrateBackupPreActivityMetadataJson`

Core fills in the default wallet id for records that lack one (or have it empty) and preserves any existing non-default id such as `trezor:{hash}`. The app only extracts and reinserts the sub-arrays via `JSONSerialization`, so it never edits Core model JSON itself. The helpers are idempotent, so current backups that already carry a wallet id are unaffected.

This is the iOS half of synonymdev/bitkit-core#113; the migration APIs it calls shipped in bitkit-core 0.4.0.

## Changes

- `BackupService.swift`: add `migrateCoreOwnedBackupFields` and apply it at the `.activity` and `.metadata` restore sites.
- Bump bitkit-core to 0.4.0 (pbxproj + Package.resolved).

## Notes

- Base is `chore/bump-bitkit-core-0.3.9` (#617), which introduces the wallet-scoped restore path this builds on. Retarget to `master` once #617 merges.


piotr-iohk (approved):
tACK

* In-place upgrade 2.3.1 → head: wallet state intact, Trezor pair + TO_SPENDING worked
* 2.3.1 VSS backup → wipe → restore on head: legacy wallet restored cleanly; logs show `Full restore success`, `Restored 3 activities, 2 activity tags`; balances/channel/order came back (on-chain + spending), Trezor pair + TO_SPENDING worked
* no issues in logs

PR #622, opened 2026-07-10
feat(lightning): show route fee for probe results

pwltr:
### Description

Updates Bitkit to use the route fee now exposed by `ldk-node` probe events.

Needs https://github.com/synonymdev/ldk-node/pull/88

- Bumps `ldk-node` to `0.7.0-rc.53`
- Handles `routeFeeMsat` on successful and failed probe events
- Displays the returned route fee in the probing tool
- Removes the follow-up fee estimation call after successful probes
- Logs probe route fees, or `unknown` when unavailable

### Screenshot / Video

image


piotr-iohk (approved):
tACK

PR #623, opened 2026-07-14
fix: tune lightning probe routing

pwltr:
### Description

Tunes Lightning scorer fee parameters to improve route selection reliability.

- Applies custom scorer fee parameters during LDK node setup.
- Keeps unspecified scorer fields from the default LDK config.
- Stores probe failure `shortChannelId` as a string for diagnostic display.

Android counterpart: https://github.com/synonymdev/bitkit-android/pull/1052


piotr-iohk (approved):
utACK

PR #620, opened 2026-07-10
feat: support paykit receiver paths

ben-kaufman:
This PR:
1. Updates Paykit to v0.1.0-rc33 and configures the mobile app as the `bitkit/wallet` receiver.
2. Discovers and saves supported `bitkit/wallet` and `bitkit/server` receivers while keeping one visible contact per Pubky identity.
3. Scopes private links, payment details, reservations, retries, and cleanup to the exact receiver path.
4. Publishes the wallet receiver marker with the app's current public/private capabilities and removes it when Paykit sharing is disabled.
5. Refreshes receiver discovery when an existing contact is scanned or pasted, while normal send-to-contact continues to target `bitkit/wallet` explicitly.

### Description

Paykit now supports multiple apps under one Pubky identity. This change adopts that model without changing Bitkit's contact UI: a person remains one contact, while their saved SDK contact record can track multiple receiver paths.

Private links are maintained for supported receivers that advertise private capabilities. Receiving details are published independently per receiver and only when that receiver can send payments, so a `bitkit/server` receiver can be discovered and linked without receiving wallet invoices or addresses. Public contact payments and fallback remain scoped to `bitkit/wallet`.

Receiver-record failures are isolated per contact, successful delivery reports are still persisted, and rescanning an existing contact shows the validation result immediately while its receiver metadata refreshes.

No migration is included because the receiver-path Paykit state has not shipped and existing development state can be discarded.

Companion Android PR: https://github.com/synonymdev/bitkit-android/pull/1066

### Linked Issues/Tasks

- Paykit receiver paths: https://github.com/pubky/paykit-rs/pull/89
- Paykit receiver markers: https://github.com/pubky/paykit-rs/pull/94

### Screenshot / Video

N/A

#### Manual Tests

- [ ] **1.** Contacts → add a Paykit-enabled Bitkit user: one contact is saved and its `bitkit/wallet` receiver is available for payment.
- [ ] **2.** Send → Contact → select the saved contact: private payment resolves through `bitkit/wallet`; public fallback still works when private payment is unavailable.
- [ ] **3.** Existing contact → scan or paste the same Pubky key after a `bitkit/server` receiver is published: the contact remains one row and the new receiver is saved in the background.
- [ ] **4.** Contact with a `bitkit/server` marker that cannot send payments: Bitkit may maintain its private link but does not publish wallet invoices or addresses to that receiver.
[...]

PR #630, opened 2026-07-15
feat: add Paykit server accounts

ben-kaufman:
### Description

This PR builds on #620 to add Paykit Server watch-only accounts:

- Presents a dedicated watch-only consent step before the normal Pubky authorization screen.
- Requires `x-bitkit-claim=watch-only-account-v1` and the exact `/pub/paykit/v0/bitkit/server/:rw` capability, rejecting missing, unknown, duplicate, or mismatched companion parameters.
- Uses Paykit companion authorization and fails closed if companion delivery or normal authorization fails.
- Allocates monotonic BIP84 accounts, reuses the same incomplete account on retry, and exports the exact versioned 84-byte claim payload.
- Automatically names new accounts from the requesting service; users can rename them later from Settings.
- Hides Settings → Advanced → Server Accounts until the experimental Paykit UI is enabled.
- Persists and backs up account metadata, serializes lifecycle and reconciliation changes, and supports disabling or restoring LDK tracking without deleting wallet state.
- Configures additional-wallet scanning and the rolling 1,000-address lookahead.
- Deterministically derives the receiver Noise key from the wallet's BIP39 seed using domain-separated HKDF-SHA256 with version, network, and receiver-path context; securely caches and verifies it so the wallet seed remains the only backup required.
- Reuses the derived Noise key across authentication and restarts, and publishes only its receiver marker when live session access becomes available.
- Updates to Paykit `0.1.0-rc37`, LDK Node `0.7.0-rc.57`, and Bitkit Core `0.4.2`; Bitkit Core now supplies trusted extended-public-key serialization, so Base58Swift and its BigInt dependency are no longer required.

The relay claim contains only the versioned account index, address type, and serialized xpub. Local names and tracking state are not disclosed. Ordinary Pubky Auth remains unchanged.

References:

- Base PR: https://github.com/synonymdev/bitkit-ios/pull/620
- Companion Android PR: https://github.com/synonymdev/bitkit-android/pull/1084
- Server boundary ADR: https://github.com/pubky/paykit-server/blob/master/docs/ADRs/0001-minimal-paykit-server-boundaries.md#creator-setup-and-bitkit-watch-only-account-claim

### QA Notes

This slice implements ADR 0001's `watch-only-account-v1` claim, which is the currently agreed protocol. ADR 0002's proposed `watch-only-account-service-v1` delegation flow has not been agreed and is not part of the current implementation.

Reviewers can exercise the complete Bitkit-side setup and account lifecycle with this disposable test request:

```text
pubkyauth://signin?caps=/pub/paykit/v0/bitkit/server/:rw&relay=https://httprelay.pubky.app/inbox/&secret=e3t7e3t7e3t7e3t7e3t7e3t7e3t7e3t7e3t7e3t7e3s&x-bitkit-claim=watch-only-account-v1
```

Enable Paykit UI, then scan or paste the URL. It exercises watch-only consent, normal Pubky authorization, account allocation and xpub export, companion-claim delivery followed by AuthToken delivery, local activation, and Server Accounts management. The xpub can be copied from the account details and used for external address derivation, funding, restart discovery, spending, and tracking disable/re-enable testing.

The request secret is public, so this URL must only be used with a disposable regtest wallet. A Paykit Server listener or receiver harness is still required to verify requester-side relay polling, decryption, and signature verification. A staging deployment is intended once Paykit Server runtime/listener composition is available.

#### Manual Tests

- [ ] Enable Paykit UI → Settings → Advanced: Server Accounts is visible; disable Paykit UI: it is hidden.
- [ ] Scanner/paste → use the disposable setup request above → approve watch-only sharing → complete normal Pubky authorization: the account becomes active.
- [ ] Copy the xpub from Server Accounts → externally derive a nonzero receive address → send funds → restart Bitkit: the additional-wallet full scan discovers the funds.
- [ ] Spend the externally received funds from Bitkit.
- [ ] Settings → Advanced → Server Accounts → disable and re-enable tracking: the same account and balance return without allocating another account.
- [ ] Missing, duplicate, unknown, or mismatched companion parameters/capability are rejected; ordinary Pubky Auth remains unchanged.

#### Automated Checks

- `WatchOnlyAccountServiceTests.swift`
- `PubkyAuthApprovalSheetTests.swift`
- `PubkyAuthRequestTests.swift`
- `PaykitReceiverNoiseKeyStoreTests.swift`
- Paykit rc37 simulator build succeeded against LDK Node rc.57.
- The exact E2E app-build mode (`DEBUG E2E_BUILD`) succeeds with the notification target's watch-only/localization membership dependencies.
- The iOS and Android receiver-key tests share an independently verified derivation vector and cover persistence, reuse, invalid caches, and wrong-seed fail-closed behavior.
- Post-delivery normal-auth failure, retry after delivery, restart retry, activation, and wallet Paykit regressions are covered by the selected suites.
- 101 selected watch-only, Paykit, Pubky Auth, endpoint, and receiver-key tests passed after rebase.
- SwiftFormat and `git diff --check` passed.


piotr-iohk (approved):

Bitkit/Services/PubkyService.swift
```diff
@@ -916,6 +1054,121 @@ private final class PaykitSdkSessionProvider: SdkPubkySessionProvider, @unchecke
...
+        let domainBytes = Data(domain.utf8)
+        let salt = Data(SHA256.hash(data: domainBytes))
+        var prk = Data(HMAC<SHA256>.authenticationCode(for: seed, using: SymmetricKey(data: salt)))
+        defer { prk.resetBytes(in: prk.startIndex ..< prk.endIndex) }
+
+        var expandInput = Data("\(version)\0\(network)\0\(receiverPath)".utf8)
+        expandInput.append(0x01)
+        return Data(HMAC<SHA256>.authenticationCode(for: expandInput, using: SymmetricKey(data: prk)))
+    }
+}
```
jvsena42, line 1095: can be replaced with the primitive `HKDF<SHA256>.deriveKey(inputKeyMaterial:salt:info:outputByteCount: 32)`

Bitkit/Managers/ScannerManager.swift
```diff
@@ -131,6 +131,16 @@ class ScannerManager: ObservableObject {
 
         Haptics.play(.scanSuccess)
```
jvsena42, line 132: success haptic could be after `!PubkyAuthRequest.isProtocolURL(uri) ` guard

jvsena42 (approved):
https://github.com/user-attachments/assets/23f90e9b-09e4-432e-baa9-fed00b3f7d35

Approved with some nits

Bitkit/Services/PubkyService.swift
jvsena42: https://github.com/synonymdev/bitkit-ios/issues/641

Bitkit/Managers/ScannerManager.swift
jvsena42: https://github.com/synonymdev/bitkit-ios/issues/642

PR #632, opened 2026-07-16
feat: add boltz savings swap

coreyphillips:
This PR adds a swap-based path for transferring spending balance to savings, so Lightning funds can move on-chain through a Boltz reverse swap instead of only by closing a channel. It ports the equivalent Android feature (synonymdev/bitkit-android#1081) to iOS, including its later refinements.

### Description

- Adds a Boltz reverse swap flow to the transfer-to-savings journey. The confirm screen shows a live quote (amount to savings, network fee, service fee, and the amount you will receive) and lets the user proceed with a swap or fall back to closing the channel.
- The flow is gated behind a new "Enable Savings Swap" toggle under Dev Settings while its UI is being finalised, so nothing changes for users until it is switched on. Dev Settings also gets a Swaps section grouping that toggle with the existing swaps list.
- Swaps are offered on mainnet only. Boltz's testnet deployment is deprecated and regtest expects a local backend, so on every other network, and whenever the dev toggle is off, the journey behaves exactly as it did before this PR: the swipe closes the channel, with no quote, fees, slider, or extra close action, and the updates stream never starts.
- Closing a channel is the default and a priced quote is the only thing that upgrades a transfer to a swap. Without one, for any reason (unsupported network, Boltz unreachable, or an amount below the swap minimum), the swipe closes the channel rather than doing nothing. The quote fetch is bounded so a hanging Boltz request cannot leave the swipe stuck loading.
- Holds back a small reserve for Lightning routing fees so the hold invoice can actually be paid.
- Waits for the on-chain claim during the progress step. If the claim has not landed within the wait window (30s), the screen shows a "settling on-chain" state so the transfer reads as committed and completing in the background rather than as a false success.
- Surfaces an unroutable payment as a failure instead of a false "settling" state. Paying the hold invoice returns as soon as the payment is dispatched and a hold invoice never settles until Boltz claims on-chain, so the progress step now watches the node's payment-failed event for the paid invoice alongside the claim: the claim, a Boltz error, a routing failure, and the timeout race, and whichever resolves first wins.
- Resumes any pending swaps when the wallet starts and refreshes balances automatically once a swap lands on-chain. The updates stream is managed in the wallet view model: it only starts where swaps are supported and enabled, retries with backoff up to a ceiling (a paid swap needs the stream to broadcast its claim), stops with the node, and restarts on the next wallet start. The progress screen also ensures the stream is running before starting a swap.
- Claims reverse swaps as soon as their lockup hits the mempool (acceptZeroConf) instead of waiting for a confirmation, for faster settlement.
- Adds a Swaps section under Dev Settings to list swaps and claim a reverse swap, backed by a new service that wraps the Boltz module. The manual claim is a recovery tool, so it is offered for any unclaimed reverse swap that has not reached a terminal state, including a swap still marked created locally or one Boltz already reports settled while no claim transaction is recorded locally. A stalled updates stream can leave the cached status behind while the funds are already claimable on-chain, and the chain is the source of truth.
- Adds a reusable continuous amount slider component used by the swap amount controls, and an optional loading state on the swipe-to-confirm button so it cannot be swiped while a quote is still being fetched.
- Bumps bitkit-core to 0.5.3, which adds the Boltz module this flow depends on and the acceptZeroConf option on the updates stream.

### Linked Issues/Tasks

Android counterpart: synonymdev/bitkit-android#1081

### Screenshot / Video

Insert relevant screenshot / recording

#### Manual Tests (swap path needs mainnet plus Dev Settings → Enable Savings Swap)
- [ ] **1.** With a spending balance, Transfer to Savings → Savings Confirm: quote shows amount to savings, network fee, service fee, and you'll receive.
- [ ] **2.** Savings Confirm → drag the amount slider: quote recomputes for the new amount.
- [ ] **3.** Savings Confirm with a balance below the swap minimum: no quote or fees show and the swipe closes the channel instead, with the separate close action hidden.
- [ ] **4a.** Savings Confirm → swipe to confirm → Savings Progress: completes after the on-chain claim, or shows the "settling on-chain" state if the claim is still landing.
  - [ ] **4b.** After claim: savings balance reflects the swapped funds without a manual sync.
  - [ ] **4c.** Swipe to confirm with no route to Boltz (e.g. no outbound liquidity): the transfer surfaces a payment failure instead of hanging on "settling".
- [ ] **5.** `regression:` Savings Confirm → Close channel instead: falls back to the channel-close path.
- [ ] **6.** `regression:` Start the app with a pending swap: swap resumes and balance updates once it lands on-chain.
- [ ] **7a.** Dev Settings → Swaps → tap a swap → Swap Details: Claim shows for any unclaimed reverse swap that is not in a terminal state, including one still marked created locally and one already reported settled with no local claim txid, the recovery cases.
  - [ ] **7b.** Swap Details of a claimed, expired, refunded, or failed swap, or any submarine swap: no Claim button.
- [ ] **8.** `regression:` On a regtest or testnet build, Transfer to Savings → Savings Confirm: no quote, fees, slider, or close-instead action, and the swipe closes the channel as before.
- [ ] **9.** `regression:` With Dev Settings → Enable Savings Swap off, Transfer to Savings → Savings Confirm: no quote, fees, slider, or close-instead action, and the swipe closes the channel as before.
- [ ] **10.** Dev Settings → Swaps section: toggling Enable Savings Swap persists across an app restart.
#### Automated Checks
- Unit tests added: savings swap quote fee breakdown, claimable status gating with permissive recovery semantics, the swap gate (unsupported off mainnet, off without the dev flag, and the dev flag alone not enabling it off mainnet), and the transfer resolving to the channel-close path without a quote, in `BitkitTests/SavingsSwapTests.swift`.
- Ran locally against bitkit-core 0.5.3: simulator Debug build succeeds and `SavingsSwapTests` pass.
- The swap execution path (payment/claim/failure race) is not unit tested: it drives the Boltz and Lightning singletons directly, which the existing suite does not mock; it mirrors the Android fix, which is unit tested there.
- No changelog fragment: the flow is behind a dev toggle, so it is not user-facing yet.


Bitkit/ViewModels/TransferViewModel.swift
```diff
@@ -1140,6 +1156,159 @@ class TransferViewModel: ObservableObject {
...
+            let swap = try await boltzService.createReverseSwap(amountSat: amount, claimAddress: claimAddress)
+            Logger.info("Created savings transfer swap \(swap.id)", context: "TransferViewModel")
+
+            // Subscribe before paying so a claim settling faster than the payment
+            // call returns cannot be missed (events buffer from stream creation).
+            let events = boltzService.events()
+
+            // Pay the hold invoice (amount is encoded). It stays pending until Boltz
+            // locks funds on-chain and we claim them, which is the expected happy path.
+            _ = try await lightningService.send(bolt11: swap.invoice)
```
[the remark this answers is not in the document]
coreyphillips: Fixed in 2f0a6fcc. The hold-invoice payment now runs in a background task and the outcome is driven off the bounded claim wait, which runs concurrently, so the 30s window is no longer gated behind the payment. If the claim has not landed in time the screen moves to the settling state; a payment that fails to initiate is still surfaced as a failure.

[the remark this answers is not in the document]
coreyphillips: Follow-up after verifying the LDK Node semantics: `Bolt11Payment.send` returns a payment id as soon as the payment is initiated and does not block until the hold invoice settles. Settlement follows the on-chain claim, which the updates stream performs, and the outcome is delivered via events. So awaiting the payment before the bounded claim wait does not gate the 30s timeout, and there is no indefinite spin. In 98fb3ab1 I kept the direct structure (which also surfaces a failure to initiate the payment immediately) and documented the non-blocking behaviour so the ordering is not mistaken for a stuck flow. The equivalent Android path (payInvoice wraps the same non-blocking send) is correct for the same reason.

Bitkit/Views/Transfer/SavingsProgressView.swift
```diff
@@ -131,42 +137,11 @@ struct SavingsProgressView: View {
...
-
-                            // Start retrying the cooperative close for non-trusted channels
-                            transfer.startCoopCloseRetries(channels: nonTrustedChannels)
-                        }
-                    }
-                } catch {
-                    app.toast(error)
+                switch transfer.savingsTransferMode {
+                case .swap:
+                    await runSavingsSwap()
```
jvsena42, line 142: Re-entering the progress screen creates and pays a second swap.
the swap runs inside the view's .task, which SwiftUI cancels on disappear. TransferViewModel holds no in-flight guard and no result memo.

Bitkit/ViewModels/WalletViewModel.swift
```diff
@@ -413,12 +413,89 @@ class WalletViewModel: ObservableObject {
...
+            await self?.startSwapUpdatesWithRetry()
+        }
+    }
+
+    /// Open the swap updates stream so any pending LN -> onchain swaps resume and auto-claim.
+    /// Uses the wallet's current fee rate for the claim tx. Retries up to a ceiling: without the
+    /// stream a paid swap has nothing to broadcast its claim, so give up only after
+    /// `swapUpdatesMaxAttempts` and leave the next trigger to retry. Once started, bitkit-core
+    /// keeps the WebSocket alive with its own reconnect loop.
+    private func startSwapUpdatesWithRetry() async {
```
jvsena42, line 456: could `swapUpdatesTask = nil ` when the retry loop exits without success for making possible recover after giving up

Bitkit/Views/Transfer/SavingsConfirmView.swift
```diff
@@ -96,11 +118,62 @@ struct SavingsConfirmView: View {
...
         .offlineOverlay(title: t("lightning__transfer__nav_title"))
+        .task(id: totalSats) {
+            // Pull the latest node balances so a just-received payment is reflected, then
+            // present the swap fee before the user commits. Recomputed when the amount changes.
+            await wallet.syncStateAsync()
+            guard totalSats > 0 else { return }
+            await transfer.loadSavingsSwapQuote(
+                requestedSat: totalSats,
+                spendableSats: UInt64(max(0, wallet.maxSendLightningSats))
+            )
```
jvsena42, line 145: savingsSwapState.isLoading is only set inside loadSavingsSwapQuote, so during syncStateAsync() the swipe button is not in its loading state

Bitkit/ViewModels/TransferViewModel.swift
```diff
@@ -1140,6 +1166,246 @@ class TransferViewModel: ObservableObject {
...
+
+    // MARK: - Savings Swap (Boltz reverse swap)
+
+    /// Fetch swap limits, derive the adjustable amount range, and publish an initial fee quote
+    /// (defaulting to the maximum transferable) so the user sees the cost before confirming.
+    /// The confirm slider then re-prices locally via `onSwapAmountChange`. A quote is the only
+    /// thing that unlocks the swap, so every failure simply leaves it nil and the transfer
+    /// falls back to closing a channel. Skipped entirely where swaps are unsupported or the
+    /// flow is switched off in dev settings, see `BoltzService.isSwapEnabled`.
+    func loadSavingsSwapQuote(requestedSat: UInt64, spendableSats: UInt64) async {
```
jvsena42, line 1192: could add a `guard !Task.isCancelled` . fetchReverseSwapLimits's network task isn't cancellation-aware. When .task(id:) restarts, the superseded invocation still assigns savingsSwapState on completion.

Bitkit/ViewModels/TransferViewModel.swift
```diff
@@ -1140,6 +1166,246 @@ class TransferViewModel: ObservableObject {
...
+            // Subscribe before paying so a claim settling faster than the payment call returns is
+            // not missed (events buffer from stream creation).
+            let events = boltzService.events()
+
+            // Pay the hold invoice. `send` returns a payment id as soon as the HTLC is dispatched;
+            // a hold invoice never settles until Boltz claims on-chain, so this does not block on
+            // the claim. A failure to dispatch the payment throws and is surfaced immediately.
+            let paymentId = try await lightningService.send(bolt11: swap.invoice)
+
+            let result = await awaitSwapOutcome(
```
jvsena42, line 1312: Android maps `event.reason.toUserMessage(context)`. The `Event.paymentFailed` payload carries the reason on iOS too

Bitkit/ViewModels/TransferViewModel.swift
```diff
@@ -1140,6 +1166,246 @@ class TransferViewModel: ObservableObject {
...
+                }
+                return nil
+            }
+            // Lightning routing failure on the paid hold invoice.
+            group.addTask {
+                while !Task.isCancelled {
+                    if await paymentFailure.hasFailed {
+                        return .failure(message: paymentFailedMessage)
+                    }
+                    try? await Task.sleep(nanoseconds: 200_000_000)
```
jvsena42: this method polls `SwapPaymentFailureCapture.hasFailed` every 200 ms (TransferViewModel.swift:1334). A CheckedContinuation resumed from the event handler removes the loop and the latency

jvsena42 (changes requested):

Bitkit/Views/Transfer/SavingsProgressView.swift
coreyphillips: Fixed in 27d7da65. The swap execution now lives on the view model as a memoized run: `executeSavingsSwap` starts at most one task per confirm commit, and re-entry awaits that same run, so SwiftUI cancelling and re-running the progress screen's `.task` neither cancels the in-flight swap nor creates and pays a second one. A completed run replays its result on re-entry instead of re-executing. `onTransferToSavingsConfirm` clears the memo, so each deliberate swipe commit still gets a fresh run. Owning the execution outside the view task also means view cancellation can no longer cut the outcome wait short mid-flow, which previously could resolve the race through the timeout branch early.

Bitkit/ViewModels/WalletViewModel.swift
coreyphillips: Done in 27d7da65. `startSwapUpdatesWithRetry` now frees the task slot when it gives up, so the next trigger (node start or entering a swap flow) can start the stream again. The nil-out sits behind the existing `Task.isCancelled` check, so a stop/restart that already recycled the slot is not clobbered. On success the slot intentionally stays occupied: `ensureSwapUpdatesRunning` short-circuits on `swapUpdatesRunning` first, and `stopSwapUpdates` still clears both.

Bitkit/Views/Transfer/SavingsConfirmView.swift
coreyphillips: Fixed in 27d7da65. The confirm screen now calls `beginSavingsSwapQuoteLoad` before `syncStateAsync`, which puts `savingsSwapState` into its loading state behind the same `isSwapEnabled` gate (so regtest and the disabled flow keep the instant swipe as before), and `loadSavingsSwapQuote` keeps it through the fetch. The swipe button is therefore in its loading state across the whole sync plus quote window and cannot briefly land in the channel-close fallback. The `totalSats == 0` early-out moved into `loadSavingsSwapQuote` so the loading marker is always cleared.

Bitkit/ViewModels/TransferViewModel.swift
coreyphillips: Done in 27d7da65. `loadSavingsSwapQuote` now checks `Task.isCancelled` right after `fetchReverseSwapLimits` returns and bails without touching `savingsSwapState`, so a superseded `.task(id:)` invocation cannot clobber the state its replacement publishes. The `reverseSwapLimits` write sits behind the same guard.

Bitkit/ViewModels/TransferViewModel.swift
coreyphillips: Done in 27d7da65. The event handler now captures the reason from `Event.paymentFailed`, and the failure message goes through `PaymentFailureReason.userMessage(for:)`, a mirror of Android's `toUserMessage`: `recipientRejected`, `retriesExhausted`, `routeNotFound` and `paymentExpired` map to their dedicated strings, and everything else (including a nil reason) falls back to the generic payment-failed description. The three strings Android has that iOS lacked were added to the English table with the same source text so Transifex keeps them aligned. Unit tests cover the mapping and the fallbacks.

Bitkit/ViewModels/TransferViewModel.swift
coreyphillips: Done in 27d7da65. `SwapPaymentFailureCapture` now parks the racing wait on a `CheckedContinuation` that the event handler resumes, so the failure branch fires the moment the node reports it and the 200 ms poll is gone. When another branch of the race wins, the group's cancellation resumes the wait as `.cancelled` through `withTaskCancellationHandler`, so the task group cannot hang on an unresumed continuation. The outcome is memoized, so a late waiter and a failure event arriving after cancellation are both handled deterministically. Unit tests cover resume-on-failure, the memo, and cancellation.

jvsena42 (approved):
Good to merge. The feature is safely hidden under dev flag. 
The user facing UI will be land on design V62

PR #637, opened 2026-07-22
feat: support incoming paykit requests

ben-kaufman:
### Description

This PR builds on #630 to support incoming Paykit payment requests:

- Updates Paykit to `0.1.0-rc39` and uses its separate public and private payment resolution APIs.
- Receives supported one-time Bitcoin payment requests while Bitkit is active, discarding expired or unsupported requests.
- Reuses the existing send confirmation and payment flow, with a Payment Request title, fixed requested amount, and requesting contact.
- After the user approves, validates that the request remains pending, consumes the private payment list, accepts the request, then submits payment.
- Uses public payment details only when no Noise channel is established and never falls back from private to public resolution.
- Persists consumed private payment-list versions before submission, treats submitted, pending, and uncertain payments as consumed, waits for newer details, and serializes private payments per contact and receiver path.
- Bounds failed presentation retries, restores pending requests after preparation failures, and refreshes immediately when connectivity returns.
- Advertises payment-request support with the live Paykit session.

Payment proofs and receipts remain out of scope.

### Linked Issues/Tasks

- Base PR: #630
- Companion Android PR: https://github.com/synonymdev/bitkit-android/pull/1098

### Screenshot / Video

N/A — the existing payment UI is reused, and no media is attached.

#### Manual Tests

- [ ] **1.** Receive a request from a linked Paykit contact while Bitkit is active: the Payment Request confirmation opens with the contact and requested amount.
- [ ] **2.** Swipe to pay: the private list is consumed before request acceptance, then payment continues through the existing send flow.
- [ ] **3.** Receive or retain a request past its expiration: it is discarded and not presented.
- [ ] **4.** Attempt another private payment before the contact publishes a newer payment list: Bitkit waits and does not reuse the consumed details or fall back to public details.

#### Automated Checks

- `PaykitPaymentRequestServiceTests.swift`: request mapping, expiry, lifecycle, consume-before-accept ordering, bounded presentation retry, preparation recovery, and presentation serialization.
- `PrivatePaykitServiceTests.swift`: consumed-list persistence and prevention of private payment-detail reuse.
- `PublicPaykitServiceTests.swift`: separate public payment endpoint parsing and resolution.
- 21 focused payment-request tests passed in the iOS simulator after the review fixes.
- SwiftFormat, translation validation, and `git diff --check` passed.


Bitkit/Views/Wallets/Send/SendConfirmationView.swift
```diff
@@ -501,53 +504,29 @@ struct SendConfirmationView: View {
                 // native millisatoshi precision instead of our truncated satoshi value.
                 let paymentSats: UInt64? = invoice.amountSatoshis == 0 ? amount : nil
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in ee9a95c8. Incoming requests now reject a fixed-amount Bolt11 whose encoded amount differs from the amount the user approved, before request acceptance, private-list consumption, or payment submission. Amountless and exact-match invoices remain supported, with focused coverage.

Bitkit/Views/Wallets/Send/LnurlPayConfirm.swift
```diff
@@ -194,6 +195,9 @@ struct LnurlPayConfirm: View {
         let contactPublicKey = app.contactPaymentContext?.publicKey
 
         do {
+            try await acceptIncomingPaymentRequest()
+            try await consumePrivatePaymentListIfNeeded()
+
             // Fetch the Lightning invoice from LNURL
             let bolt11 = try await LnurlHelper.fetchLnurlInvoice(
```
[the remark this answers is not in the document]
ben-kaufman: This ordering is intentional. Invoking the private LNURL callback uses the endpoint, and rc39 requires any submitted, pending, or uncertain use to consume the entire Private Payment List before that callback so the same private details cannot be reused. Accepting the request records the user approval; settlement proofs and receipts are outside this PR.

Bitkit/Services/PrivatePaykitService+Backup.swift
```diff
@@ -18,7 +30,11 @@ extension PrivatePaykitService {
         state = PrivatePaykitState(contacts: [:])
         knownSavedContactKeys.removeAll()
         if let backup {
-            try await PaykitSdkService.shared.restoreBackupState(backup)
+            let decoded = try JSONDecoder().decode(Backup.self, from: Data(backup.utf8))
+            try await PaykitSdkService.shared.restoreBackupState(decoded.sdkState)
```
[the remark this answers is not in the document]
ben-kaufman: No code change here: this raw backup format only exists in the unshipped parent Paykit work, so there are no production backups to migrate. Per the pre-release scope of this stack, we are intentionally not adding migration or backward-compatibility code.

Bitkit/Services/PaykitPaymentRequestService.swift
```diff
@@ -0,0 +1,367 @@
...
+        let counterpartyReceiverPath: String
+    }
+
+    let paymentRequestId: String
+    let counterparty: String
+    let counterpartyReceiverPath: String
+    let amountValue: String
+    let amountSats: UInt64
+    let paymentReference: String
+    let expiresAt: Date?
```
jvsena42, line 16: is this intended to be optional? just double checking

Bitkit/AppScene.swift
```diff
@@ -699,6 +725,100 @@ struct AppScene: View {
...
+        guard !requests.isEmpty else { return }
+
+        isPresentingPaykitPaymentRequest = true
+        defer { isPresentingPaykitPaymentRequest = false }
+
+        for request in requests {
+            do {
+                let result = try await PrivatePaykitService.shared.beginPaymentRequest(request)
+                guard sheets.activeSheetConfiguration == nil, app.contactPaymentContext == nil else { return }
+                guard case let .opened(paymentTarget, privatePaymentContext) = result else { continue }
```
jvsena42: if beginPaymentRequest never returns `.opened` (the counterparty has no private payment list, or no accepted endpoint matches) this method takes the continue branch and `markPresentedIfPending` is never called. That request then re-runs full SDK resolution on every 30 s poll and on every sheet dismissal, indefinitely, with no backoff and no user-visible feedback

jvsena42: There is also no reject path

Bitkit/AppScene.swift
```diff
@@ -699,6 +725,100 @@ struct AppScene: View {
...
+
+        while !Task.isCancelled {
+            do {
+                try await Task.sleep(for: Self.paykitPaymentRequestRefreshInterval)
+            } catch {
+                return
+            }
+            await refreshIncomingPaykitPaymentRequests()
+        }
+    }
```
jvsena42, line 758: This method perform multiple requests every 30 seconds for as log the app is in foreground, even if nothings has changed

Maybe should back off the requests when nothing has been returned or scaling the interval by linked-peer count

Bitkit/Services/PrivatePaykitService+Payments.swift
```diff
@@ -69,12 +117,67 @@ extension PrivatePaykitService {
...
+
+        switch linkState {
+        case nil, .notLinked, .linking:
+            return true
+        case .linked, .recoveryRequired, .blocked, .unknown:
+            return false
+        }
+    }
+
+    private func hasLiveSessionForCurrentProfile() async throws -> Bool {
```
jvsena42: this method has no remaining callers. Dead code?

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
```diff
@@ -501,53 +510,29 @@ struct SendConfirmationView: View {
...
-                    }
                     throw error
                 }
             } else if app.selectedWalletToPayFrom == .onchain, let invoice = app.scannedOnchainInvoice {
                 let amount = wallet.sendAmountSats ?? invoice.amountSatoshis
                 let useMaxAmount = await shouldUseMaxOnchainSend(address: invoice.address, amountSats: amount)
+                try await consumePrivatePaymentListIfNeeded(contactPaymentContext)
                 let txid = try await wallet.send(address: invoice.address, sats: amount, isMaxAmount: useMaxAmount)
-                if let contactPublicKey {
-                    await PrivatePaykitService.shared.discardRemoteOnchainEndpoints(
```
jvsena42, line 546: This was the last caller of this method. Is it dead code?

jvsena42 (changes requested):

Bitkit/Services/PaykitPaymentRequestService.swift
ben-kaufman: Yes. The SDK allows proposalExpiresAt to be absent, so the app keeps non-expiring requests actionable. When an expiration is supplied, malformed or already-expired values are rejected during mapping, and live requests are removed as soon as their expiration is reached.

Bitkit/Services/PrivatePaykitService+Payments.swift
ben-kaufman: Yes. Removed in 801e6982; hasLiveSessionForCurrentProfile had no remaining callers.

Bitkit/AppScene.swift
ben-kaufman: Correct. The explicit reject path will come with the remaining incoming-payment-request discovery and action design. For now, dismissing the reused payment sheet leaves the request pending rather than accepting or rejecting it.

ben-kaufman: Addressed in 801e6982. Failed presentation now uses per-request retry backoff of 30, 60, 120, then 300 seconds, including sheet dismissal and setup failures, so the same unresolved request no longer runs full resolution on every poll.

Bitkit/AppScene.swift
ben-kaufman: Addressed in 801e6982. Foreground synchronization now backs off from 30 to 60 to 120 seconds while the actionable request set is unchanged, and resets to 30 seconds whenever requests change.

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
ben-kaufman: Yes for the on-chain helper: its last caller was removed, so discardRemoteOnchainEndpoints was removed in 801e6982. discardRemoteLightningEndpoints remains because reusablePrivateEndpoints still calls it when filtering stale Bolt11 entries.

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
```diff
@@ -483,9 +486,15 @@ struct SendConfirmationView: View {
...
     private func performPayment() async throws {
         var createdMetadataPaymentId: String? = nil
-        let contactPublicKey = app.contactPaymentContext?.publicKey
+        let contactPaymentContext = app.contactPaymentContext
+        let contactPublicKey = contactPaymentContext?.publicKey
 
         do {
+            try validateIncomingPaymentRequestContext(contactPaymentContext)
+            try validateIncomingPaymentRequestAmounts(contactPaymentContext)
+            try await acceptIncomingPaymentRequest()
```
jvsena42: **D1 — accept/consume ordering is inverted vs Android (blocking).**

```
Android  proceedWithPayment:  validate → consume → accept → send
iOS      performPayment:      validate → accept  → validate → consume → send
```

Because `accept()` runs first here, if `consumePrivatePaymentListIfNeeded` then throws `paymentListAlreadyConsumed` (lines 513 / 534), you land in a state where:

1. the counterparty has been told the request is **accepted**,
2. `PaykitPaymentRequestManager.perform` has already removed the request from `pendingRequests`, and
3. **no payment is ever sent**.

Android's ordering makes that unreachable — the consume failure short-circuits before `acceptIncomingPaymentRequestIfNeeded` is called. Suggest swapping to consume-then-accept to match.

Note `LnurlPayConfirm.swift` has the same inverted order and needs the same fix.

Bitkit/Services/PaykitPaymentRequestService.swift
```diff
@@ -0,0 +1,396 @@
...
+            !presentedRequestIds.contains($0.id) && (presentationRetryDates[$0.id].map { $0 <= date } ?? true)
+        }
+    }
+
+    func deferPresentation(_ request: PaykitPaymentRequest) {
+        discardExpiredRequests()
+        guard pendingRequests.contains(where: { $0.id == request.id }) else { return }
+
+        let attempt = presentationRetryAttempts[request.id, default: 0]
+        let delay = Self.presentationRetryDelays[min(attempt, Self.presentationRetryDelays.count - 1)]
```
jvsena42: **D2 — no retry cap; iOS retries forever.**

`min(attempt, count - 1)` clamps to the *longest* delay rather than terminating, so once a request has been deferred 4 times it keeps retrying every 300s for the life of the session.

Android stops permanently after exhausting the list, via two mechanisms:
- the presentation filter: `retryAttempts <= PAYKIT_PAYMENT_REQUEST_PRESENTATION_RETRY_DELAYS.size`
- `deferPaymentRequestPresentation`: `val retryDelay = DELAYS.getOrNull(attempt) ?: return`

So the new backoff bounds the *frequency* on iOS but not the *total* — an unresolvable request (e.g. a counterparty that publishes no private payment list) still re-runs full SDK resolution indefinitely. Suggest treating an out-of-range `attempt` as "give up" and filtering those requests out of `requestsForPresentation()`.

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
```diff
@@ -501,53 +510,29 @@ struct SendConfirmationView: View {
                 // native millisatoshi precision instead of our truncated satoshi value.
                 let paymentSats: UInt64? = invoice.amountSatoshis == 0 ? amount : nil
                 do {
+                    try await consumePrivatePaymentListIfNeeded(contactPaymentContext)
```
jvsena42: **D3 — no recovery when payment preparation fails.**

If this consume (or the `accept` above) throws, the error propagates out of `performPayment` and nothing re-defers the request. It stays in `presentedRequestIds`, so `requestsForPresentation()` filters it out and it is **never re-presented for the rest of the session** — even though it is still pending and still payable once the counterparty publishes a newer list.

Android handles this explicitly in `handlePaymentPreparationFailure`:

```kotlin
toast(error)
val request = context?.incomingPaymentRequest
if (request != null && paykitPaymentRequestRepo.isPending(request)) {
    presentedPaymentRequestIds.remove(request.id)
    deferPaymentRequestPresentation(request)
}
```

Worth mirroring — combined with D2, an iOS user who hits a transient preparation failure loses the request entirely until they restart the app.

Bitkit/AppScene.swift
```diff
@@ -700,6 +726,115 @@ struct AppScene: View {
         }
     }
 
+    @discardableResult
+    private func refreshIncomingPaykitPaymentRequests() async -> Bool {
```
jvsena42, line 729: **D4 — no connectivity-triggered refresh.**

Android refreshes as soon as the device reconnects:

```kotlin
private fun observePaykitPaymentRequestConnectivity() {
    viewModelScope.launch {
        isOnline.drop(1)
            .filter { it == ConnectivityState.CONNECTED }
            .collect { refreshIncomingPaykitPaymentRequests() }
    }
}
```

iOS has no equivalent hook despite `NetworkMonitor` being available, so after regaining connectivity a pending request waits for the next poll tick — up to 120s once the interval has backed off. An `.onChange` on the network monitor calling `refreshIncomingPaykitPaymentRequests()` would close the gap and also reset the poll backoff index naturally.

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
```diff
@@ -590,6 +575,47 @@ struct SendConfirmationView: View {
...
+
+        let paymentAmount = switch app.selectedWalletToPayFrom {
+        case .lightning:
+            wallet.sendAmountSats ?? app.scannedLightningInvoice?.amountSatoshis
+        case .onchain:
+            wallet.sendAmountSats ?? app.scannedOnchainInvoice?.amountSatoshis
+        }
+
+        guard let paymentAmount, request.acceptsPaymentAmount(paymentAmount) else {
+            throw LnurlPayInvoiceMismatchError()
```
jvsena42: **D5 — misleading error type for payment-request mismatches.**

`LnurlPayInvoiceMismatchError` is thrown here (and at 599 / 604) for plain onchain and Lightning payment-request amount mismatches that have nothing to do with LNURL. It surfaces through the generic send-failure path, so the user — and anyone triaging a bug report — sees an LNURL error for a non-LNURL flow.

Android added a dedicated string and handler:

```kotlin
private fun rejectMismatchedPaymentRequest() {
    toast(
        type = Toast.ToastType.ERROR,
        title = context.getString(R.string.wallet__toast_payment_failed_title),
        description = context.getString(R.string.wallet__payment_request_mismatch),
        testTag = "PaymentFailedToast",
    )
    hideSheet()
}
```

Suggest a `PaykitPaymentRequestError.amountMismatch` case plus the matching `wallet__payment_request_mismatch` string so the two platforms report the same condition the same way.

(Android's extra `isPaymentRequest && incomingPaymentRequest == null` sanity guards aren't needed here — iOS reads `app.contactPaymentContext?.incomingPaymentRequest` directly rather than keeping a separate `SendUiState.isPaymentRequest` copy that could desync.)

Bitkit/Services/PaykitPaymentRequestService.swift
```diff
@@ -0,0 +1,396 @@
...
+        let counterparty: String
+        let counterpartyReceiverPath: String
+    }
+
+    let paymentRequestId: String
+    let counterparty: String
+    let counterpartyReceiverPath: String
+    let amountValue: String
+    let amountSats: UInt64
+    let paymentReference: String
```
jvsena42: **D6 — `paymentReference` and `metadata` are parsed but never read.**

Both are populated in `init?(record:now:)` (lines 59 and 62) and then never referenced anywhere in the app — the only occurrences in the codebase are the declarations and the assignments. Android's `PaykitPaymentRequest` omits both entirely.

Bitkit/AppScene.swift
```diff
@@ -700,6 +726,115 @@ struct AppScene: View {
...
+            let requestsChanged = await refreshIncomingPaykitPaymentRequests()
+            if requestsChanged {
+                refreshIntervalIndex = 0
+            } else {
+                refreshIntervalIndex = min(refreshIntervalIndex + 1, Self.paykitPaymentRequestRefreshIntervals.count - 1)
+            }
+        }
+    }
+
+    private func presentNextIncomingPaykitPaymentRequest() async {
```
jvsena42, line 760: **D7 — the presentation loop is untested because it lives in a View.**

Android's `AppViewModelSendFlowTest.kt` covers exactly this orchestration layer: presentation, polling, approval ordering, expiry, and duplicate/pending consumption — including `verify(privatePaykitRepo).consumePrivatePaymentList(...)` ordering assertions. iOS test coverage stops at `PaykitPaymentRequestManager`; everything in this method is untested.

Worth noting **D1 is precisely the class of bug those ordering assertions catch**.

Moving this loop into `PaykitPaymentRequestManager` (or a small dedicated coordinator) would make it testable and would also resolve the `isPresentingPaykitPaymentRequest`-as-`@State` smell — that flag is a re-entrancy lock on manager state that currently lives in view state. It works today only because the guard-and-set pair happens synchronously on the MainActor, which is a fragile invariant to rely on.

jvsena42 (changes requested):
Reviewed against the companion Android PR (synonymdev/bitkit-android#1098). The two implementations are impressively well-aligned on the layers that matter most — identical backoff constants (refresh `[30,60,120]s`, presentation retry `[30,60,120,300]s`), identical request-mapping rules, equivalent amount parsing despite very different implementations, matching network scoping, and the same `canUsePublicPayment` / `WaitingForUpdatedPaymentList` policy. The follow-up backoff commit also cleanly resolved the unbounded retry loop and the dead code from the previous round.

Seven divergences from Android remain, flagged inline as **D1–D7**. **D1** is the one I'd block on: iOS and Android invert the accept/consume ordering, and the iOS order has a reachable state where the counterparty is told "accepted" but no payment is sent. D2 and D3 are follow-ups to the same backoff commit's intent; D4–D7 are alignment polish.

Verified on `c4f32ae6`: `xcodebuild` Debug (iPhone 16 Pro sim) **BUILD SUCCEEDED**, SwiftFormat clean on all changed files.

---

**Not blocking, carried over from the previous round:** no reject action and `presentedRequestIds` is in-memory only (so a non-expiring request re-presents on every app launch — same on Android, so it reads as a deliberate shared scope decision); `beginSavedContactPayment` can block a tap for ~12s across its retry loop; redundant `wallet.sendAmountSats` write at `AppScene.swift:824` that `SendSheet.onAppear` immediately overwrites; and 15 hand-edited locale files where the repo convention is English-only plus a Transifex sync.

Bitkit/Services/PaykitPaymentRequestService.swift
jvsena42: probably will be used in future PR?

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
ben-kaufman: Fixed in b380b328. Payment preparation is now one manager operation that consumes the private list before accepting the request, then removes the pending request only after both steps succeed. Both regular and LNURL send flows use that operation, with ordering covered by a regression test.

Bitkit/Services/PaykitPaymentRequestService.swift
ben-kaufman: Fixed in b380b328. Presentation now allows the initial attempt plus the four configured delayed retries, logs once when the schedule is exhausted, and then excludes that request from further presentation attempts while normal synchronization continues.

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
ben-kaufman: Fixed in b380b328. A non-cancellation failure while consuming or accepting refreshes manager state, removes the presented marker for a still-pending request, and schedules it through the bounded retry path. The recovery behavior has focused test coverage.

Bitkit/AppScene.swift
ben-kaufman: No code change was needed here: AppScene.handleNetworkChange already calls refreshIncomingPaykitPaymentRequests when connectivity becomes true, so reconnection triggers an immediate refresh. I verified the existing hook and left a single implementation in place.

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
ben-kaufman: Fixed in b380b328. Payment-request amount mismatches now throw PaykitPaymentRequestError.amountMismatch and use the same dedicated user-facing message as Android, rather than reporting an LNURL-specific error.

Bitkit/Services/PaykitPaymentRequestService.swift
ben-kaufman: Fixed in b380b328. Removed paymentReference and metadata from the app model and mapping because the current UI does not consume them. They can be introduced with the later request-details design if needed.

Bitkit/AppScene.swift
ben-kaufman: Addressed in b380b328 without adding a separate coordinator. Request selection and presentation reentrancy now live in PaykitPaymentRequestManager, the SwiftUI-only navigation work remains in AppScene, and tests cover presentation serialization, retry exhaustion, preparation recovery, and consume-before-accept ordering.

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
```diff
@@ -483,9 +486,14 @@ struct SendConfirmationView: View {
...
     private func performPayment() async throws {
         var createdMetadataPaymentId: String? = nil
-        let contactPublicKey = app.contactPaymentContext?.publicKey
+        let contactPaymentContext = app.contactPaymentContext
+        let contactPublicKey = contactPaymentContext?.publicKey
 
         do {
+            try validateIncomingPaymentRequestContext(contactPaymentContext)
+            try validateIncomingPaymentRequestAmounts(contactPaymentContext)
+            try await prepareIncomingPaymentRequest()
```
jvsena42, line 495: **N1 — the post-accept ownership re-check was dropped in this refactor.**

The previous version re-validated after the accept round-trip:

```swift
try await acceptIncomingPaymentRequest()
try validateIncomingPaymentRequestContext(contactPaymentContext)   // ← removed
```

That call (`:575-579`) checked **two** things — expiry *and* `app.ownsContactPaymentContext(context)`. `perform()` re-checks expiry and pending-ness internally before running the operation, so expiry is still covered. But nothing re-checks **context ownership** after the awaits, and the invoice is read from `app` *after* them:

```swift
try await prepareIncomingPaymentRequest()          // consume + accept, network round-trips
if app.selectedWalletToPayFrom == .lightning, let invoice = app.scannedLightningInvoice {
```

Concrete trigger: `perform` mutates `pendingRequests` on success, which fires the `.onChange` at `AppScene.swift:230`. If the request expires at that instant, that handler runs `app.resetSendState()` + `wallet.resetSendState(...)` **mid-submission** — clearing the context and `sendAmountSats` while `performPayment` is still executing, after consume and accept have already been committed and persisted.

Android guards exactly this case with a submission flag:

```kotlin
if (
    !isSubmittingPaymentRequest &&
    currentSheet.value is Sheet.Send &&
    requests.none { it.id == activeRequest.id }
) { hideSheet() }
```

The window is narrow, but it's a money path where the private payment list has already been consumed and the counterparty has already been told "accepted". Cheapest fix: re-add `try validateIncomingPaymentRequestContext(contactPaymentContext)` immediately after this line. More robust: an iOS equivalent of `isSubmittingPaymentRequest` gating the `AppScene.swift:230` handler.

`LnurlPayConfirm.swift:200` has the same shape and needs the same treatment.

Bitkit/Services/PaykitPaymentRequestService.swift
```diff
@@ -0,0 +1,433 @@
...
+    case requestUnavailable
+    case requestExpired
+    case operationInProgress
+    case amountMismatch
+
+    var errorDescription: String? {
+        switch self {
+        case .amountMismatch:
+            t("wallet__payment_request_mismatch")
+        case .requestUnavailable, .requestExpired, .operationInProgress:
```
jvsena42: **N2 — three of the four cases now render as debug text to the user.**

Adopting `LocalizedError` gave `.amountMismatch` a real message (👍 for D5), but these three return `nil`. All three are reachable from `perform()` — `.requestExpired` when the request lapses between presentation and swipe, `.requestUnavailable` when a concurrent refresh drops it, `.operationInProgress` on a double-invocation — and they propagate up to `app.toast(error)` (`AppViewModel.swift:295-300`), which uses `error.localizedDescription`.

With a `nil` `errorDescription`, Foundation falls back to:

> The operation couldn't be completed. (Bitkit.PaykitPaymentRequestError error 1.)

`.requestExpired` is the realistic one to hit — swipe to pay just after a short-lived request lapses, and that's what you get.

Android carries a message on every case:

```kotlin
sealed class PaykitPaymentRequestError(message: String) : AppError(message) {
    data object RequestUnavailable : PaykitPaymentRequestError("Payment request is unavailable")
    data object RequestExpired : PaykitPaymentRequestError("Payment request has expired")
}
```

Same class of divergence D5 just fixed, one layer down — suggest localized strings for all four cases (or at minimum `.requestExpired` and `.requestUnavailable`, since `.operationInProgress` is arguably an internal condition that shouldn't surface as a toast at all).

jvsena42 (changes requested):
Round 3 — reviewed `b380b328` against the previous round. Good progress: **D1, D2, D3, D5, D6 are properly fixed**, and the new tests target the actual behaviour rather than restating it (`testPreparationConsumesBeforeAccepting` asserting ordering from inside the consume closure is a nice touch, as is `testDeferredRequestStopsAfterConfiguredRetries` walking the full 30/60/120/300s ladder to the cap). The redundant `wallet.sendAmountSats` write in `AppScene` is gone too.

**D7 is partially addressed** — the re-entrancy flag moved into the manager as `isPresentingRequests` with `presentRequests(_:)`, so the `@State` smell is resolved and covered by `testPresentationOperationIsNotReentered`; the resolve → claim → `handleScannedData` → route → `markPresented` loop itself still lives untested in `AppScene`. **D4 (connectivity-triggered refresh) is not addressed.**

Two new issues introduced by the refactor are flagged inline as **N1** and **N2**. N1 is the one I'd want handled before merge: consolidating the accept/consume into `prepareForPayment` dropped a post-await guard that the manager structurally cannot replace.

Verified on `b380b328`: `xcodebuild` Debug (iPhone 16 Pro sim) **BUILD SUCCEEDED**, SwiftFormat clean on all 6 changed files. I could not run `scripts/validate-translations.js` locally (`Cannot find module 'glob'` — my environment, not this branch), so the new `wallet__payment_request_mismatch` key is unverified by that check; adding it to `en.lproj` only is the right call per repo convention.

Bitkit/Services/PaykitPaymentRequestService.swift
ben-kaufman: Fixed in `2f78dbc4`. `requestUnavailable`, `requestExpired`, and `operationInProgress` now each have localized English fallback messages, so every `PaykitPaymentRequestError` produces a user-facing description. I also added a focused regression test covering all four cases; translation validation passes.

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
ben-kaufman: Fixed in `2f78dbc4`. Both the direct invoice/on-chain flow and the LNURL flow now revalidate the captured payment-request context immediately after `prepareForPayment` returns, before reading payment state or starting the payment. This catches expiry or ownership loss caused while consume/accept is awaiting.

jvsena42 (dismissed):
utAck, implementation still not testable end-to-end

jvsena42 (approved):

PR #638, opened 2026-07-22
refactor: batch paykit contact cleanup

ben-kaufman:
### Description

This PR builds on #637 to address the Paykit cleanup performance follow-ups from #620:

- Reads the SDK linked-peer list once per multi-contact publication or cleanup pass, with one shared retry instead of a per-contact fallback.
- Clears receiver-scoped payment lists as before, then drains outbound private messages and checks pending status once for the whole successful batch.
- Preserves per-contact failure isolation: contacts with failed or still-pending cleanup retain their cached state and retry marker, while successful contacts are cleared.
- Serializes cleanup with endpoint publication and skips the local state wipe if relevant contact state changes during an awaited SDK operation.

On the error path, a failed link snapshot still clears locally recorded published receiver paths while retaining the contact state and retry marker. A batch-level drain-inspection failure conservatively keeps the whole affected batch pending for retry. No user-facing behavior changes are intended.

### Linked Issues/Tasks

- Base PR: #637
- Review follow-up: #620
- Companion Android PR: https://github.com/synonymdev/bitkit-android/pull/1099

### Screenshot / Video

N/A — service-only refactor.

#### Manual Tests

N/A

#### Automated Checks

- `PrivatePaykitServiceTests.swift`: all 13 focused tests passed in the iOS simulator, including mixed cleanup success/failure coverage.
- SwiftFormat passed for the changed Swift files.
- `git diff --check` passed.

### DIFF
--- a/Bitkit/Services/PrivatePaykitService+Contacts.swift
+++ b/Bitkit/Services/PrivatePaykitService+Contacts.swift
@@ -87,53 +87,66 @@ extension PrivatePaykitService {
     }
 
     func removePublishedEndpoints(for publicKeys: [String]) async throws {
-        var firstError: Error?
-        var didChangeState = false
-        for publicKey in normalizedSavedContactKeys(publicKeys) {
-            var didFail = false
-            let cleanupReceiverPaths: [String]
-            do {
-                cleanupReceiverPaths = try await receiverPathsForCleanup(publicKey: publicKey)
-            } catch {
-                firstError = firstError ?? error
-                Self.markDeletedContactCleanupPending([publicKey])
-                continue
-            }
+        let publicKeys = normalizedSavedContactKeys(publicKeys)
+        guard !publicKeys.isEmpty else { return }
 
+        let linkedReceiverPaths: [String: Set<String>]
+        do {
+            linkedReceiverPaths = try await linkedReceiverPathsByPublicKey()
+        } catch {
+            Self.markDeletedContactCleanupPending(publicKeys)
+            throw error
+        }
+
+        var firstError: Error?
+        var failedPublicKeys = Set<String>()
+        var clearedRetryKeys = [PrivateMessageDrainRetryKey]()
+        var didChangeState = false
+        for publicKey in publicKeys {
+            let cleanupReceiverPaths = receiverPathsForCleanup(
+                publicKey: publicKey,
+                linkedReceiverPaths: linkedReceiverPaths[publicKey, default: []]
+            )
             for receiverPath in cleanupReceiverPaths {
                 do {
                     let report = try await PaykitSdkService.shared.clearPrivatePaymentList(to: publicKey, receiverPath: receiverPath)
                     if !report.failedToQueue.isEmpty || !report.failedToDeliver.isEmpty {
                         throw PrivatePaykitError.privateUnavailable
                     }
-                    let retryKey = PrivateMessageDrainRetryKey(publicKey: publicKey, receiverPath: receiverPath)
-                    await drainPendingPrivateMessages(reason: "cleanup", advancing: [retryKey])
-                    if await pendingPrivateMessageDrainKeys([retryKey]).contains(retryKey) {
-                        throw PrivatePaykitError.privateUnavailable
-                    }
+                    clearedRetryKeys.append(PrivateMessageDrainRetryKey(publicKey: publicKey, receiverPath: receiverPath))
                 } catch {
-                    didFail = true
+                    failedPublicKeys.insert(publicKey)
                     firstError = firstError ?? error
-                    Self.markDeletedContactCleanupPending([publicKey])
                 }
-            }
-
-            if !didFail {
-                if var contactState = state.contacts[publicKey] {
-                    let hadStateToClear = !contactState.cachedResolvedEndpoints.isEmpty ||
-                        !contactState.localInvoicesByReceiverPath.isEmpty ||
-                        !contactState.publishedPrivatePaymentReceiverPaths.isEmpty
-                    contactState.cachedResolvedEndpoints = []
-                    contactState.localInvoicesByReceiverPath = [:]
-                    contactState.publishedPrivatePaymentReceiverPaths = []
-                    let shouldRemoveContact = !contactState.hasCacheState
-                    state.contacts[publicKey] = shouldRemoveContact ? nil : contactState
-                    didChangeState = didChangeState || hadStateToClear || shouldRemoveContact
-                }
-
-                Self.clearDeletedContactCleanupPending([publicKey])
             }
         }
+
+        if !clearedRetryKeys.isEmpty {
+            await drainPendingPrivateMessages(reason: "cleanup", advancing: clearedRetryKeys)
+            let pendingRetryKeys = await pendingPrivateMessageDrainKeys(clearedRetryKeys)
+            if !pendingRetryKeys.isEmpty {
+                failedPublicKeys.formUnion(pendingRetryKeys.map(\.publicKey))
+                firstError = firstError ?? PrivatePaykitError.privateUnavailable
+            }
+        }
+
+        let successfulPublicKeys = Set(publicKeys).subtracting(failedPublicKeys)
+        for publicKey in successfulPublicKeys {
+            if var contactState = state.contacts[publicKey] {
+                let hadStateToClear = !contactState.cachedResolvedEndpoints.isEmpty ||
+                    !contactState.localInvoicesByReceiverPath.isEmpty ||
+                    !contactState.publishedPrivatePaymentReceiverPaths.isEmpty
+                contactState.cachedResolvedEndpoints = []
+                contactState.localInvoicesByReceiverPath = [:]
+                contactState.publishedPrivatePaymentReceiverPaths = []
+                let shouldRemoveContact = !contactState.hasCacheState
+                state.contacts[publicKey] = shouldRemoveContact ? nil : contactState
+                didChangeState = didChangeState || hadStateToClear || shouldRemoveContact
+            }
+        }
+
+        Self.markDeletedContactCleanupPending(Array(failedPublicKeys))
+        Self.clearDeletedContactCleanupPending(Array(successfulPublicKeys))
 
         if didChangeState {
             persistState(markWalletBackup: true)
@@ -265,6 +278,16 @@ extension PrivatePaykitService {
         var firstError: Error?
         var updates = [PrivatePaymentListReservationUpdateInput]()
         var linkRetryKeys = [PrivateMessageDrainRetryKey]()
+        let linkedReceiverPaths: [String: Set<String>]
+        let linkedReceiverPathsError: Error?
+        do {
+            linkedReceiverPaths = try await linkedReceiverPathsByPublicKey()
+            linkedReceiverPathsError = nil
+        } catch {
+            linkedReceiverPaths = [:]
+            linkedReceiverPathsError = error
+        }
+
         for publicKey in publicKeys {
             let receiverPaths: [String]
             do {
@@ -296,17 +319,18 @@ extension PrivatePaykitService {
                 )
             }
             let cleanupReceiverPaths: [String]
-            do {
-                cleanupReceiverPaths = try await receiverPathsForPrivateEndpointCleanup(
-                    publicKey: publicKey,
-                    excluding: publicationReceiverPaths + receiverPathSelection.cleanupProtectedReceiverPaths
-                )
-            } catch {
+            if let linkedReceiverPathsError {
                 cleanupReceiverPaths = []
-                firstError = firstError ?? error
+                firstError = firstError ?? linkedReceiverPathsError
                 Logger.warn(
-                    "Failed to inspect private Paykit links for \(PubkyPublicKeyFormat.redacted(publicKey)) during \(reason): \(error)",
+                    "Failed to inspect private Paykit links for \(PubkyPublicKeyFormat.redacted(publicKey)) during \(reason): \(linkedReceiverPathsError)",
                     context: "PrivatePaykit"
+                )
+            } else {
+                cleanupReceiverPaths = receiverPathsForPrivateEndpointCleanup(
+                    publicKey: publicKey,
+                    excluding: publicationReceiverPaths + receiverPathSelection.cleanupProtectedReceiverPaths,
+                    linkedReceiverPaths: linkedReceiverPaths[publicKey, default: []]
                 )
             }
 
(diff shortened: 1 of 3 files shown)


Bitkit/Services/PrivatePaykitService+Contacts.swift
```diff
@@ -293,18 +316,19 @@ extension PrivatePaykitService {
...
             }
             let cleanupReceiverPaths: [String]
-            do {
-                cleanupReceiverPaths = try await receiverPathsForPrivateEndpointCleanup(
-                    publicKey: publicKey,
-                    excluding: publicationReceiverPaths + receiverPathSelection.cleanupProtectedReceiverPaths
-                )
-            } catch {
+            if let linkedReceiverPathsError {
                 cleanupReceiverPaths = []
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in be606b00. The normal path still reads `linkedPeers()` once, but if that shared lookup fails we now retry per contact. If an individual fallback also fails, we still build cleanup updates and retry keys from the locally recorded published paths, while preserving that contact's error.

Bitkit/Services/PrivatePaykitService+Contacts.swift
```diff
@@ -87,54 +87,86 @@ extension PrivatePaykitService {
...
+            let pendingRetryKeys = await pendingPrivateMessageDrainKeys(clearedRetryKeys)
+            if !pendingRetryKeys.isEmpty {
+                failedPublicKeys.formUnion(pendingRetryKeys.map(\.publicKey))
+                firstError = firstError ?? PrivatePaykitError.privateUnavailable
+            }
+        }
 
-                Self.clearDeletedContactCleanupPending([publicKey])
+        let successfulPublicKeys = Set(publicKeys).subtracting(failedPublicKeys)
+        for publicKey in successfulPublicKeys {
```
jvsena42: **Actor-reentrancy window for the state wipe is now much wider.**

`PrivatePaykitService` is an `actor` (reentrant), and `removePublishedEndpoints` is *not* run under `withPublicationLock` — only `syncLocalEndpointPublication` is. Before this PR each contact's `state.contacts[X]` wipe happened immediately after that contact's own clear+drain. Now every wipe is deferred to here, after the whole batch's clears *and* the full drain.

Scenario: `removePublishedEndpoints()` (the no-arg variant, which unions all known + cached + pending keys) runs while a foreground `refreshKnownSavedContactEndpoints` is in flight. Contact A is cleared early in the batch; during the many subsequent awaits the concurrent publish calls `recordPublishedPrivatePaymentList` for A (or an incoming invoice writes `localInvoicesByReceiverPath`); this loop then wipes `cachedResolvedEndpoints` / `localInvoicesByReceiverPath` / `publishedPrivatePaymentReceiverPaths` for A, desyncing local state from what the SDK actually holds.

The hazard pre-exists, but the batching widens it from one-contact latency to whole-batch latency. Cheapest mitigations: snapshot the relevant `state.contacts[publicKey]` fields before the clear and skip the wipe if they changed, or run the cleanup pass under `withPublicationLock` as well.

Bitkit/Services/PrivatePaykitService+Contacts.swift
```diff
@@ -87,54 +87,86 @@ extension PrivatePaykitService {
...
-                firstError = firstError ?? error
-                Self.markDeletedContactCleanupPending([publicKey])
-                continue
+        for publicKey in publicKeys {
+            let contactLinkedReceiverPaths: Set<String>
+            if let linkedReceiverPathsSnapshot {
+                contactLinkedReceiverPaths = linkedReceiverPathsSnapshot[publicKey, default: []]
+            } else {
+                do {
+                    contactLinkedReceiverPaths = try await linkedReceiverPaths(publicKey: publicKey)
```
jvsena42: **This fallback re-issues the exact call that just failed.**

`linkedReceiverPaths(publicKey:)` now calls `linkedReceiverPathsByPublicKey()`, which is the same `PaykitSdkService.shared.linkedPeers()` that just threw when the snapshot was taken. So when the snapshot fails, this pass makes **N+1** `linkedPeers()` calls instead of the pre-PR N — a regression in exactly the dimension the PR optimizes — and the "retrying per contact" wording in the warning above oversells what the retry can accomplish.

Suggest picking one: on snapshot failure either fail the pass once (set `firstError`, mark all keys pending, return) or continue with an empty snapshot, and drop the per-contact re-fetch entirely. Same applies to the mirrored block in `syncLocalEndpointPublicationLocked`.

Bitkit/Services/PrivatePaykitService+Contacts.swift
```diff
@@ -295,20 +338,26 @@ extension PrivatePaykitService {
...
-                Logger.warn(
-                    "Failed to inspect private Paykit links for \(PubkyPublicKeyFormat.redacted(publicKey)) during \(reason): \(error)",
-                    context: "PrivatePaykit"
-                )
+            let contactLinkedReceiverPaths: Set<String>
+            if let linkedReceiverPathsSnapshot {
+                contactLinkedReceiverPaths = linkedReceiverPathsSnapshot[publicKey, default: []]
+            } else {
+                do {
+                    contactLinkedReceiverPaths = try await linkedReceiverPaths(publicKey: publicKey)
```
jvsena42: **Duplicated snapshot-or-fallback block.**

This is the same ~14 lines as the block in `removePublishedEndpoints` (lines 109–124), differing only in the log's `reason`. Per the repo's reuse convention, worth extracting into a single helper — e.g. `linkedReceiverPaths(for publicKey: String, snapshot: [String: Set<String>]?, reason: String) async -> (paths: Set<String>, error: Error?)` — so the two call sites can't drift.

(Same redundant-refetch caveat as the other block applies here too.)

Bitkit/Services/PrivatePaykitService+Contacts.swift
```diff
@@ -87,54 +87,86 @@ extension PrivatePaykitService {
...
+                    failedPublicKeys.insert(publicKey)
+                    firstError = firstError ?? error
+                    Logger.warn(
+                        "Failed to inspect private Paykit links for \(PubkyPublicKeyFormat.redacted(publicKey)) during cleanup: \(error)",
+                        context: "PrivatePaykit"
+                    )
+                }
             }
-
+            let cleanupReceiverPaths = receiverPathsForCleanup(
```
jvsena42, line 110: **Behavior change worth calling out** (the description says "no user-facing behavior changes are intended").

Previously, when `receiverPathsForCleanup` threw for a contact, the loop `continue`d and made **zero** clear attempts for it. Now a link-inspection failure yields an empty `contactLinkedReceiverPaths` and execution falls through to here, so the contact still gets `clearPrivatePaymentList` called for its locally-published paths.

This is arguably an improvement (partial progress, and the contact is marked failed either way so it retries), but it's a real change in what hits the SDK on the error path. Probably worth a line in the PR description.

Bitkit/Services/PrivatePaykitService+Contacts.swift
```diff
@@ -87,54 +87,86 @@ extension PrivatePaykitService {
...
-                    contactState.localInvoicesByReceiverPath = [:]
-                    contactState.publishedPrivatePaymentReceiverPaths = []
-                    let shouldRemoveContact = !contactState.hasCacheState
-                    state.contacts[publicKey] = shouldRemoveContact ? nil : contactState
-                    didChangeState = didChangeState || hadStateToClear || shouldRemoveContact
-                }
+        if !clearedRetryKeys.isEmpty {
+            await drainPendingPrivateMessages(reason: "cleanup", advancing: clearedRetryKeys)
+            let pendingRetryKeys = await pendingPrivateMessageDrainKeys(clearedRetryKeys)
+            if !pendingRetryKeys.isEmpty {
```
jvsena42, line 131: **Blast radius of a transient drain-check failure grew from one contact to the whole batch.**

`pendingPrivateMessageDrainKeys` returns *all* the passed keys when `linkedPeers()` or `pendingOutboundPrivateCounterparties()` throws (see its two `catch` blocks). With the check now run once per batch, a single transient SDK hiccup marks every successfully-cleared contact as failed, so none of them get their state cleared or their retry marker cleared.

The retry path covers it, so this isn't a correctness bug — but it's a real reduction in transient-failure granularity that comes with the batching, and it's the kind of thing that shows up as "cleanup never converges" if the SDK is flaky during a full cleanup pass. Worth noting in the description alongside the perf win.

Bitkit/Services/PrivatePaykitService+Contacts.swift
```diff
@@ -87,54 +87,86 @@ extension PrivatePaykitService {
...
+                contactState.cachedResolvedEndpoints = []
+                contactState.localInvoicesByReceiverPath = [:]
+                contactState.publishedPrivatePaymentReceiverPaths = []
+                let shouldRemoveContact = !contactState.hasCacheState
+                state.contacts[publicKey] = shouldRemoveContact ? nil : contactState
+                didChangeState = didChangeState || hadStateToClear || shouldRemoveContact
             }
         }
 
+        Self.markDeletedContactCleanupPending(Array(failedPublicKeys))
```
jvsena42: **Failure markers are now written once at the end instead of eagerly.**

The old code called `markDeletedContactCleanupPending([publicKey])` at the moment each contact failed, so a crash mid-loop still left the marker persisted. Now nothing is written until the batch completes.

Mostly recoverable, since `removePublishedEndpoints()` and `pruneUnsavedContactState` re-derive keys from `state.contacts` (which isn't cleared for failed contacts). The narrow gap is a contact deleted via `removeSavedContact` that has no local cache state — its only record would be this marker, and a crash before this line loses it. Low severity, but it is a durability regression relative to the eager write.

Bitkit/Services/PrivatePaykitService+Contacts.swift
```diff
@@ -87,54 +87,86 @@ extension PrivatePaykitService {
...
+            await drainPendingPrivateMessages(reason: "cleanup", advancing: clearedRetryKeys)
+            let pendingRetryKeys = await pendingPrivateMessageDrainKeys(clearedRetryKeys)
+            if !pendingRetryKeys.isEmpty {
+                failedPublicKeys.formUnion(pendingRetryKeys.map(\.publicKey))
+                firstError = firstError ?? PrivatePaykitError.privateUnavailable
+            }
+        }
 
-                Self.clearDeletedContactCleanupPending([publicKey])
+        let successfulPublicKeys = Set(publicKeys).subtracting(failedPublicKeys)
```
jvsena42: Minor: `publicKeys` is already deduplicated by `normalizedSavedContactKeys` at the top of the function, so this `Set(publicKeys)` is rebuilding a set that could be constructed once above the loop and reused (it's also implicitly the membership source for the loop). Purely cosmetic.

jvsena42 (commented):
Reviewed the batching refactor. The per-contact failure isolation contract does hold — I traced the partial-failure cases (one path of two fails, drain still pending, link inspection fails) and each ends with the contact in `failedPublicKeys`, state untouched, retry marker set. SwiftFormat lint passes and omitting a changelog fragment is correct for a `refactor:` with no user-facing change.

Comments inline. The one I'd want addressed (or explicitly accepted) before merge is the widened actor-reentrancy window on the deferred state wipe; the rest are smaller.

**Test coverage:** no test changes here. `PrivatePaykitServiceTests.swift` has 12 tests, none of which touch `removePublishedEndpoints` or `linkedPeers`, so the new batching, the `failedPublicKeys`/`successfulPublicKeys` split, and the snapshot-failure fallback are all untested. The "all 10 focused tests passed" note in the description is a regression run, not coverage of this change. Given the state-clearing logic is the risky part, a test asserting "contact A succeeds, contact B fails → A cleared + marker cleared, B state retained + marker set" would be worth adding.

**Merge order:** base is `codex/paykit-incoming-payment-requests`, so this needs to land behind #637.

Bitkit/Services/PrivatePaykitService+Contacts.swift
ben-kaufman: Addressed in 073bb5dd. Cleanup now uses the same publication lock as endpoint publication, snapshots the cleanup-owned contact fields before SDK awaits, and defers the local wipe while retaining the retry marker if those fields changed during the operation.

Bitkit/Services/PrivatePaykitService+Contacts.swift
ben-kaufman: Addressed in 073bb5dd. Deleted-contact cleanup markers are now persisted synchronously in removeSavedContact and removeSavedContacts before the first remote await, closing the no-cache crash gap. Successful cleanup clears only the corresponding markers.

Bitkit/Services/PrivatePaykitService+Contacts.swift
ben-kaufman: Addressed in 073bb5dd. The per-contact fallback is gone. Both cleanup and publication now use one shared linked-peer snapshot helper that makes one initial call and at most one batch-wide retry, so the failure path is bounded at two calls rather than N+1.

Bitkit/Services/PrivatePaykitService+Contacts.swift
ben-kaufman: Agreed. I kept the partial-progress behavior intentionally: after both shared link-snapshot attempts fail, locally recorded published paths are still cleared, while the contact state and retry marker are retained. The PR description now calls this error-path behavior out explicitly.

Bitkit/Services/PrivatePaykitService+Contacts.swift
ben-kaufman: Addressed in 073bb5dd. The normalized keys are converted to publicKeySet once and that set is reused for failed/successful membership calculations.

Bitkit/Services/PrivatePaykitService+Contacts.swift
ben-kaufman: Agreed on the tradeoff. This remains deliberately conservative because a failed batch-level inspection cannot prove which clears have fully drained. The affected contacts keep their state and retry markers, so the next cleanup pass retries safely. The PR description now calls out this batch-level failure granularity.

Bitkit/Services/PrivatePaykitService+Contacts.swift
ben-kaufman: Addressed in 073bb5dd. The duplicated snapshot/fallback blocks are replaced by the shared linkedReceiverPathsSnapshot helper, used by both cleanup and publication.

ovitrif (approved):
Looks good to merge. Prior cleanup/batching findings are addressed at 907b4ac, and I did not find new blocking issues in the shared linked-peer snapshot, publication-lock cleanup, or per-contact failure isolation.

PR #652, opened 2026-08-04
feat(send): improve Lightning send failure recovery

### LINKED ISSUE #483: Improve Send Error Sheet UX
> I fixed the toast message in this PR but the error sheet message is also misleading, I will remove in next PR, then on this error screen we'll have parity with iOS:

That screen needs work on both platforms. Design has some improvements to status quo but needs to be adjusted somewhat. For LN payments "Transaction" is not accurate, should be "Payment". And we should try to show a user actionable error description as well as we can.

image

_Originally posted by @pwltr in https://github.com/synonymdev/bitkit-android/issues/825#issuecomment-4014633605_

pwltr:
## Summary

- Move failed send payments onto the failure screen with localized user-facing messages.
- Add a `Try Again` action that resets Lightning routing caches only for routing-related failures, then retries the original send flow.
- Avoid repeated routing-cache resets after one reset attempt in the same send flow.
- Consolidate `PaymentFailureReason` user-message mapping into `PaymentFailureReason+UserMessage.swift` with generic and send-specific contexts.
- Fix stale Lightning node lifecycle state where the view model could think the node was running after the service node was gone.
- Add a Support action on send failures that opens a prefilled support report with failure type, payment method, routing reset status, and the Lightning payment request when available.
- Keep non-routing retries lightweight by returning to the original confirmation flow without resetting routing data.
- Fix the swipe-to-confirm control so it clears its loading state after handled send failures.

Closes https://github.com/synonymdev/bitkit-ios/issues/483

Normal Payment:

https://github.com/user-attachments/assets/f1f1922b-6b09-48fe-aa2b-84f3551aa53b

QuickPay:

https://github.com/user-attachments/assets/368cf5e4-3f1b-4a90-a129-13770ff719b1


Bitkit/Resources/Localization/ru.lproj/Localizable.strings
```diff
@@ -992,7 +992,18 @@
...
 "wallet__toast_payment_failed_title" = "Платеж не выполнен";
-"wallet__toast_payment_failed_description" = "Ваш мгновенный платеж не удался. Пожалуйста, попробуйте снова.";
+"wallet__payment_failed_description" = "Ваш мгновенный платеж не удался. Пожалуйста, попробуйте снова.";
+"wallet__payment_recipient_rejected" = "Получатель отклонил этот платеж Lightning. Проверьте счет и попробуйте снова.";
+"wallet__payment_abandoned" = "Платеж Lightning был остановлен до завершения.";
+"wallet__payment_expired" = "Срок действия этого платежа Lightning истек. Запросите новый счет.";
+"wallet__payment_timeout" = "Время ожидания платежа истекло. Попробуйте еще раз.";
+"wallet__send_payment_route_not_found" = "Bitkit не смог найти маршрут Lightning для этого платежа. Нажмите Повторить, чтобы обновить маршрутизацию платежей и найти новый путь.";
+"wallet__send_payment_retries_exhausted" = "Bitkit попробовал несколько маршрутов Lightning, но платеж не удалось завершить. Нажмите Повторить, чтобы обновить маршрутизацию платежей и попробовать новый путь.";
+"wallet__payment_route_not_found" = "Bitkit не смог найти маршрут Lightning для этого платежа.";
```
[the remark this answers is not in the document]
pwltr: No � replacement characters are present in the localization files.

Bitkit/Views/Wallets/Send/SendFailure.swift
```diff
@@ -35,4 +69,37 @@ struct SendFailure: View {
...
+
+            do {
+                var cacheResetError: Error?
+                do {
+                    try await wallet.resetPaymentRoutingCaches()
+                } catch {
+                    cacheResetError = error
+                }
+
+                try await wallet.start()
```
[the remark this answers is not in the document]
pwltr: Fixed in [2261ea2](https://github.com/synonymdev/bitkit-ios/pull/652/commits/2261ea26e7c9d70c71937de71d6b69cdeea2b280)

Bitkit/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -1305,11 +1305,18 @@
...
-"wallet__toast_payment_failed_recipient_rejected" = "The recipient rejected this payment. Try a different amount.";
-"wallet__toast_payment_failed_retries_exhausted" = "Could not find a route with sufficient liquidity. Try a smaller amount or wait and try again.";
-"wallet__toast_payment_failed_route_not_found" = "Could not find a payment path to the recipient.";
+"wallet__payment_failed_description" = "Your instant payment failed. Please try again.";
+"wallet__payment_recipient_rejected" = "The recipient rejected this Lightning payment. Please check the invoice and try again.";
+"wallet__payment_abandoned" = "The Lightning payment was stopped before it completed.";
+"wallet__payment_expired" = "This Lightning payment expired. Please request a new invoice.";
+"wallet__payment_timeout" = "Payment timed out. Please try again.";
+"wallet__send_payment_route_not_found" = "Bitkit couldn't find a Lightning route for this payment. Tap Retry to update payment routing and check for a new path.";
+"wallet__send_payment_retries_exhausted" = "Bitkit tried several Lightning routes, but the payment could not be completed. Tap Retry to update payment routing and try a new path.";
```
jvsena42, line 1314: IDK if this message would improve the user experience 🤔

Bitkit/Views/Wallets/Send/SendFailure.swift
```diff
@@ -35,4 +69,37 @@ struct SendFailure: View {
...
+
+        Task { @MainActor in
+            defer {
+                wallet.isRetryingLightningPayment = false
+            }
+
+            do {
+                var cacheResetError: Error?
+                do {
+                    try await wallet.resetPaymentRoutingCaches()
```
jvsena42, line 144: retry is wiping the cache even for failures nor related to routing, like on-chain (edge-case) , poor internet or an invoice that expired mid payment

https://github.com/user-attachments/assets/0f964686-2d94-49cf-b3f4-44f077439df7

jvsena42: Worst case is poor internet, because the app would have to download  the full graph + scorer on slow connection

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
```diff
@@ -473,6 +442,41 @@ struct SendConfirmationView: View {
...
+                }
+            } else {
+                let shouldProceed = await requestPinCheck()
+                guard shouldProceed else {
+                    throw CancellationError()
+                }
+            }
+        }
+
+        try await performPayment()
```
jvsena42, line 478: In addition to https://github.com/synonymdev/bitkit-ios/pull/652/changes#r3757164032 , there is also the case of an app with healthy routing data that simply can find a route at that moment, because of the invoice amount, poorly connected destination, low routing liquidity, etc. All these cases likely return `routeNotFound`.

For this scenario, the app would wipe health data, wait for download it again and possibly retry the same paths.

Last time I tried to improve this, I couldn't find a good way to differentiate a legit failure from stale routing data

jvsena42 (changes requested):

Bitkit/Resources/Localization/en.lproj/Localizable.strings
pwltr: yeah I'm also not 100% happy with this wording. Any suggestions?

Bitkit/Views/Wallets/Send/SendFailure.swift
pwltr: agree, the only relevant failure cases warranting reset of routing caches are `.routeNotFound` and `.retriesExhausted`. The other cases should just return to the correct confirmation screen without cache reset.

Bitkit/Views/Wallets/Send/SendConfirmationView.swift
pwltr: In this case there is not much we can do client side, at that point it is a LSP and/or network issue. Maybe a good addition would be to show a "contact support" button if the failure persists after first retry. That would help us look into insufficient LSP channel liquidity or bad scorer data for the payment target.

pwltr: Fetching RGS and remote scorer data is fairly cheap (I think <10MB) so I wouldn't worry to much about detecting healthy data. IMO it is fine to have this as a sort of 'nuke cache' option.

Bitkit/Resources/Localization/en.lproj/Localizable.strings
pwltr: I removed the additional "Tap Retry to update payment routing and check for a new path." as it was not adding much value for the UX.

jvsena42 (approved):
Ack

PR #653, opened 2026-08-04
fix: prepare Paykit local auth and private links

ben-kaufman:
### Description

- require Paykit Server watch-only authorization requests to include both public and private server capabilities, independent of capability ordering
- update to Paykit `0.1.0-rc43` and use one cached Pubky client configuration for session bootstrap and the Paykit SDK
- route local E2E Electrum, Homegate, and Pubky/Paykit through one documented `E2E_LOCAL_HOST` setting while keeping production defaults unchanged
- refresh saved contacts from Pubky so receiver paths published after contact creation, including `bitkit/server`, are discovered and persisted
- run a bounded two-second private link/message and Payment Request polling burst after contact discovery, foregrounding, and network restoration, then return to the existing low-frequency cadence
- preserve private-only payment resolution with no public-payment fallback

### Linked Issues/Tasks

- Local test environment fixes: https://github.com/pubky/locks/pull/11
- Paykit SDK local-testnet configuration and rc43 release: https://github.com/pubky/paykit-rs/pull/130
- Matching Paykit Server retry behavior: https://github.com/pubky/paykit-server/pull/2
- Matching Android behavior: https://github.com/synonymdev/bitkit-android/pull/1141

### Validation

- Paykit `0.1.0-rc43` resolves from its published tag and the app builds against its published XCFramework
- `PaykitSdkClientConfigTests` and `PrivatePaykitServiceTests`: 16/16 passed after rebasing onto current `master`
- SwiftFormat and whitespace checks passed
- Clean Locks → Paykit Server → Bitkit proof is being rerun for both platforms

### Screenshot / Video

N/A — no new UI; this changes local E2E configuration and private-link synchronization.


Bitkit/Info.plist
```diff
@@ -21,6 +21,8 @@
 	</array>
 	<key>E2E_BACKEND</key>
 	<string>$(E2E_BACKEND)</string>
+	<key>E2E_HOMEGATE_URL</key>
+	<string>$(E2E_HOMEGATE_URL)</string>
 	<key>E2E_NETWORK</key>
 	<string>$(E2E_NETWORK)</string>
```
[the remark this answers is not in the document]
ben-kaufman: Xcode expands undefined build-setting placeholders in the processed Info.plist to an empty string. I verified the built app has `E2E_HOMEGATE_URL = ""`, and `infoPlistValue` returns `nil` for that value, so the `http://127.0.0.1:6288` fallback is used. No code change is needed here.

Bitkit/Models/PubkyAuthRequest.swift
```diff
@@ -5,7 +5,13 @@ enum PubkyAuthClaim: String, Equatable {
     case watchOnlyAccountV1 = "watch-only-account-v1"
 
     static let queryParameter = "x-bitkit-claim"
-    static let watchOnlyAccountCapabilities = "/pub/paykit/v0/bitkit/server/:rw"
+    static let watchOnlyAccountCapabilities = "/pub/paykit/v0/bitkit/server/:rw,/pub/paykit/v0/private/bitkit/server/:rw"
```
ovitrif, line 8: This now requires both public and private Paykit Server capabilities for a watch-only claim, but PubkyAuthApprovalSheetTests still call PubkyAuthRequest.parse with only `/pub/paykit/v0/bitkit/server/:rw` plus the claim. Those fixtures therefore throw `invalidBitkitClaimCapabilities` before the consent/approval assertions run, so the approval-sheet suite fails on this change. Could we update those watch-only auth URL fixtures to use both capabilities from `PubkyAuthClaim.watchOnlyAccountCapabilities`?

ovitrif (changes requested):
Requesting changes for two issues after this watch-only capability change:
- PubkyAuthApprovalSheetTests still build public-only watch-only auth URLs, so PubkyAuthRequest.parse throws before the consent and approval assertions run.
- Docs/watch-only-account-claim-v1.md is incorrect because it still documents only the public server capability as the exact watch-only requirement.

Bitkit/Models/PubkyAuthRequest.swift
ben-kaufman: Fixed in 30868138. Every watch-only approval-sheet fixture now uses one shared auth URL helper built from PubkyAuthClaim.watchOnlyAccountCapabilities, so the public and private permissions stay aligned with the production contract. The focused request and approval-sheet suites pass all 50 tests on rc43.

ovitrif (dismissed):
Looks good. The previously missing private Paykit Server capability is now required in the approval-sheet fixtures and watch-only claim contract.

jvsena42 (commented):
Starting review

Bitkit.xcodeproj/project.pbxproj
```diff
@@ -1175,7 +1175,7 @@
 			repositoryURL = "https://github.com/pubky/paykit-rs";
 			requirement = {
 				kind = exactVersion;
-				version = "0.1.0-rc39";
+				version = "0.1.0-rc43";
```
jvsena42, line 1178: nit: version doesn't match PR description

Bitkit/Services/PubkyService.swift
```diff
@@ -925,7 +926,15 @@ actor PaykitSdkService {
...
 
     private func bootstrap() throws -> PubkySessionBootstrap {
-        try PubkySessionBootstrap()
+        try PubkySessionBootstrap.withPubkyClientConfig(pubkyClient: Self.pubkyClientConfig())
+    }
+
+    nonisolated static func pubkyClientConfig(isLocalE2EBackend: Bool = Env.isLocalE2EBackend) -> PubkyClientConfig {
+        var config = Paykit.defaultPubkyClientConfig()
+        if isLocalE2EBackend {
+            config.localTestnetHost = "localhost"
```
jvsena42: worth porting `E2E_LOCAL_HOST`  from android, feeding Homegate, Electrum and Paykit, plus .env.example/README docs

Bitkit/Constants/Env.swift
```diff
@@ -71,7 +71,7 @@ enum Env {
         (infoPlistValue("E2E_BACKEND") ?? "local").lowercased()
     }
 
-    private static var isLocalE2EBackend: Bool {
+    static var isLocalE2EBackend: Bool {
```
jvsena42: can be private again if implement https://github.com/synonymdev/bitkit-ios/pull/653/changes#r3728560169

Bitkit/Services/PubkyService.swift
```diff
@@ -925,7 +926,15 @@ actor PaykitSdkService {
     }
 
     private func bootstrap() throws -> PubkySessionBootstrap {
-        try PubkySessionBootstrap()
+        try PubkySessionBootstrap.withPubkyClientConfig(pubkyClient: Self.pubkyClientConfig())
+    }
+
+    nonisolated static func pubkyClientConfig(isLocalE2EBackend: Bool = Env.isLocalE2EBackend) -> PubkyClientConfig {
+        var config = Paykit.defaultPubkyClientConfig()
```
jvsena42: `pubkyClientConfig` isn't cached, `Paykit.defaultPubkyClientConfig()` is a native ffi call that fires on every `bootstrap` call, plus each `handle`

Android use `by lazy`

Bitkit/Info.plist
```diff
@@ -21,6 +21,8 @@
 	</array>
 	<key>E2E_BACKEND</key>
 	<string>$(E2E_BACKEND)</string>
+	<key>E2E_HOMEGATE_URL</key>
+	<string>$(E2E_HOMEGATE_URL)</string>
```
jvsena42, line 25: could document this in `README.md`, `AGENTS.md`, or `Docs/AI_DEVICE_TESTS.md`, where `E2E_BACKEND`/`E2E_NETWORK` are documented.

jvsena42 (changes requested):
No issues found,, only some missing cleanup ports from Android

Bitkit.xcodeproj/project.pbxproj
ben-kaufman: Updated the PR description to rc43, matching the project and resolved package versions.

Bitkit/Services/PubkyService.swift
ben-kaufman: Implemented in 59f77630. One E2E_LOCAL_HOST now feeds Electrum, Homegate, and the Pubky/Paykit local testnet; the simulator default remains 127.0.0.1 and physical-device builds can use the Mac's LAN address.

Bitkit/Constants/Env.swift
ben-kaufman: Implemented in 59f77630. isLocalE2EBackend is private again; Paykit receives only the derived optional local-testnet host.

Bitkit/Services/PubkyService.swift
ben-kaufman: Implemented in 59f77630. The Pubky client config is now constructed once per PaykitSdkService and reused by both SDK creation and session bootstrap, so the native default-config call is no longer repeated.

Bitkit/Info.plist
ben-kaufman: Documented in README.md in 59f77630 alongside E2E_BACKEND and E2E_NETWORK, including simulator defaults, the physical-device override, and E2E_HOMEGATE_URL.

jvsena42:
LGTM, waiting for CI to pass before approving

jvsena42 (approved):

PR #654, opened 2026-08-05
chore: add just clean modules recipe

jvsena42:
### Description

Fixes #647

This PR:
1. Adds a `just clean modules` recipe that clears Xcode's explicit precompiled module caches
2. Adds a README Troubleshooting section covering this failure and the two adjacent SPM failures that produce the same symptom
3. Points AGENTS.md at it from the build commands section

Bumping any of the Rust/UniFFI dependencies changes the size of the generated FFI header. Xcode keys its explicit precompiled modules on that header's size and mtime, and in explicit-modules mode a stale module is a hard error, because nothing is allowed to rebuild it mid-compile. The build then fails with a size-mismatch error followed by a long cascade of `cannot find 'uniffi_bitkitcore_checksum_func_*' in scope`.

That cascade is the whole problem. It is identical to what a genuinely stale SPM binary artifact produces, so the instinct is to wipe the package caches and re-download a few hundred megabytes, when clearing a few module directories would have fixed it in under a minute. Xcode's Reset Package Caches does not clear them, which is why the heavier fix looks like the only one that works.

The recipe goes on the existing `just clean` dispatcher rather than into a new script, so it reuses the `remove_path` guard already there. It covers both DerivedData roots the repo uses, the shared `ModuleCache.noindex` that `MODULE_CACHE_DIR` actually points at, and the indexer's own copy under `Index.noindex` — the last of these is not in the issue's workaround, and going stale there gives phantom errors in the editor without breaking the build.

The issue also proposed setting `SWIFT_ENABLE_EXPLICIT_MODULES = NO`. That would make this class of error disappear, but every CI workflow here runs on ephemeral runners that cache SPM and never DerivedData, so every CI run is a clean build — precisely the case explicit modules exists to speed up. It would also trade a loud error for silently stale modules, and leave the genuinely-stale-artifact case untouched. It is documented as a one-off `xcodebuild` override instead, with the reasoning, so the next person who reaches for it finds the answer rather than the question.

No app code is touched, so there is no changelog fragment.

### Linked Issues/Tasks

Fixes #647

### Screenshot / Video

N/A — tooling and documentation only.

#### Manual Tests

> Reproducing this needs two revisions whose FFI headers differ in size. `master` and `feat/send-swap` will not work: bitkit-core 0.5.2 and 0.5.3 ship byte-identical headers (128544 bytes, same sha). `origin/chore/dependabot-npm-cleanup` is a one-dependency delta at bitkit-core 0.4.2, header 121271 bytes.

- [x] **1.** Build on `chore/dependabot-npm-cleanup` (bitkit-core 0.4.2) to warm the cache → checkout `master` (0.5.3) → resolve → rebuild: fails with `file '...bitkitcoreFFI.h' has been modified since the module file '...bitkitcoreFFI-<hash>.pcm' was built`, `note: size changed from expected 121271 to 128544`, and 13 `cannot find checksum func` errors.
  - [x] **1b.** Reverse direction (0.5.3 warm → 0.4.2): same failure mirrored, `size changed from expected 128544 to 121271`.
- [x] **2.** `just clean modules` → rebuild: BUILD SUCCEEDED, 28s. Reverse direction: 37s.
- [x] **3.** After `just clean modules` → inspect DerivedData: `Build/Products`, `SourcePackages` and every `*.build` intermediate survive; only the four module directories and `ModuleCache.noindex` are removed.
- [x] **4a.** `just clean bogus`: prints the usage line to stderr, exits 1.
  - [x] **4b.** `just clean module-cache` alias, and `just clean modules spm` combined: both work.
  - [x] **4c.** `just clean modules` run twice in a row: idempotent, exits 0.
- [x] **5.** `just` and `just list`: show `clean [build|derived-data|modules|spm|all]...`.
- [x] **6.** README's symbol-check grep, run verbatim: returns 1 for a symbol present in the resolved artifact and 0 for a 0.4.2-only symbol, so it distinguishes the two failure modes as documented.

#### Automated Checks

- No Swift or app code changed, so no test coverage was added, modified or removed.
- Full Debug build of the `Bitkit` scheme passes on the iPhone 16 simulator after the change.
- Two of the documented failure modes were confirmed by hitting them during this work rather than by inspection: the first CLI `-resolvePackageDependencies` dying with `fatalError` / `file not found at path: ....xcframework.zip`, and an aborted resolve leaving an artifact half-extracted with only a `__MACOSX` directory, which re-resolving reports as successful without repairing.
- CI: standard checks run by the PR bot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,67 @@ This installs a pre-commit hook that lints Swift files with SwiftFormat.
 
 Due to the Rust dependencies in the project, Xcode previews are only compatible with iOS 17 and below.
 
+### Common Tasks
+
+Day-to-day commands live in the `Justfile`:
+
+```bash
+# Build, install and launch on a connected device (wraps ./run.sh)
+just run [release] [logs]
+
+# Clear caches, cheapest first
+just clean [build|derived-data|modules|spm|all]...
+```
+
+## Troubleshooting
+
+### After bumping a Rust/UniFFI dependency
+
+`bitkit-core`, `ldk-node`, `paykit` and `vss-rust-client-ffi` ship UniFFI-generated headers whose size changes on almost every version. Xcode keys its explicit precompiled modules (`.pcm`) on the header's size and mtime, and in explicit-modules mode a stale `.pcm` is a hard error — nothing is allowed to rebuild it mid-compile. The build fails with:
+
+```
+error: file '.../include/bitkitcoreFFI.h' has been modified since the module file '.../bitkitcoreFFI-<hash>.pcm' was built
+note: size changed from expected 121271 to 128544
+```
+
+followed by a cascade of misleading `cannot find 'uniffi_bitkitcore_checksum_func_*' in scope` errors in `bitkitcore.swift`. The root-cause line is easy to miss — it is emitted as `<unknown>:0:` and sorts away from the rest. Clear the module caches and rebuild:
+
+```bash
+just clean modules
+```
+
+Xcode's **File ▸ Packages ▸ Reset Package Caches** does *not* clear these. Note that `vss-rust-client-ffi` tracks `master`, so this can bite after a plain re-resolve with no version change in the diff.
+
+### `cannot find checksum func` with no "has been modified" error
+
+Same message, different cause: the downloaded binary artifact genuinely lacks the symbol, because a re-cut tag reuses a URL that SPM has already cached. Check which one you have before reaching for the heavier fix:
+
+```bash
+grep -c uniffi_bitkitcore_checksum_func_<name> \
+  ~/Library/Developer/Xcode/DerivedData/Bitkit-*/SourcePackages/artifacts/bitkit-core/BitkitCoreFFI/BitkitCore.xcframework/ios-arm64-simulator/Headers/bitkitcoreFFI.h
+```
+
+Non-zero → module cache problem, `just clean modules` is enough. Zero → the artifact really is stale:
+
+```bash
+just clean spm
+xcodebuild -resolvePackageDependencies -project Bitkit.xcodeproj -scheme Bitkit
+```
+
+The first CLI resolve often dies with `fatalError` or `file not found at path: .../<Name>.xcframework.zip` while fetching the large binary artifacts. Run it again — it can take two or three attempts.
+
+### `There is no XCFramework found at ...`
+
+An aborted resolve can leave an artifact half-extracted, with only a `__MACOSX` directory where the `.xcframework` should be. Re-running `-resolvePackageDependencies` reports success without repairing it, because SPM's workspace state still claims the artifact is present. Use `just clean spm` and re-resolve.
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -114,6 +114,67 @@ This installs a pre-commit hook that lints Swift files with SwiftFormat.
...
+Xcode's **File ▸ Packages ▸ Reset Package Caches** does *not* clear these. Note that `vss-rust-client-ffi` tracks `master`, so this can bite after a plain re-resolve with no version change in the diff.
+
+### `cannot find checksum func` with no "has been modified" error
+
+Same message, different cause: the downloaded binary artifact genuinely lacks the symbol, because a re-cut tag reuses a URL that SPM has already cached. Check which one you have before reaching for the heavier fix:
+
+```bash
+grep -c uniffi_bitkitcore_checksum_func_<name> \
+  ~/Library/Developer/Xcode/DerivedData/Bitkit-*/SourcePackages/artifacts/bitkit-core/BitkitCoreFFI/BitkitCore.xcframework/ios-arm64-simulator/Headers/bitkitcoreFFI.h
+```
```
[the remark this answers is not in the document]
jvsena42: fixed in https://github.com/synonymdev/bitkit-ios/pull/654/commits/b97f8b093be6758e52005adfc3c81b660c4f2fbc

README.md
```diff
@@ -114,6 +114,74 @@ This installs a pre-commit hook that lints Swift files with SwiftFormat.
...
+note: size changed from expected 121271 to 128544
+```
+
+followed by a cascade of misleading `cannot find 'uniffi_bitkitcore_checksum_func_*' in scope` errors in `bitkitcore.swift`. The root-cause line is easy to miss — it is emitted as `<unknown>:0:` and sorts away from the rest. Clear the module caches and rebuild:
+
+```bash
+just clean modules
+```
+
+Xcode's **File ▸ Packages ▸ Reset Package Caches** does *not* clear these. Note that `vss-rust-client-ffi` tracks `master`, so this can bite after a plain re-resolve with no version change in the diff.
```
pwltr: if this is posing a problem we should pin the package using `version` instead of `branch` and remove this transient note.

pwltr (dismissed):
I can recommend https://sweetpad.hyzyla.dev/ as an alternative to Xcode. It basically does the same as your `Justfile` plus has additional helpers.

Also note that the `README.md` is getting a bit too noisy for a user facing repo. Should probably split it into `DEVELOPMENT.md` in a follow up

README.md
jvsena42: fixed in https://github.com/synonymdev/bitkit-ios/pull/654/commits/36ed9a3757c4bf68904d84ea9309ed28055e27b5 , https://github.com/synonymdev/bitkit-ios/pull/654/commits/3d3fde5cdb128f5748f3d8eb4fedeb7f0128c786

pwltr (approved):

PR #663, opened 2026-08-12
fix(send): preserve pending send failure details

pwltr:
## Summary

- Preserve the LDK `PaymentFailureReason` when a Lightning payment times out into Pending and later fails.
- Pass pending payment context through to the failure screen so it shows the correct localized failure message.
- Keep routing-cache reset retry eligibility intact for pending failures like `routeNotFound` and `retriesExhausted`.
- Carry the BOLT11 through the pending route so the support report prefill still has payment context.

Follow-up to https://github.com/synonymdev/bitkit-ios/pull/652, no changelog entry needed.

https://github.com/user-attachments/assets/271de387-cb60-43eb-8ec0-fed637d41d15

### DIFF
--- a/Bitkit/Views/Wallets/Send/SendPendingScreen.swift
+++ b/Bitkit/Views/Wallets/Send/SendPendingScreen.swift
@@ -26,6 +26,8 @@ struct HourglassLoadingView: View {
 struct SendPendingScreen: View {
     let paymentHash: String
     let retryRoute: SendRetryRoute
+    let paymentRequest: String?
+    let routingCacheResetAttempted: Bool
     @Binding var navigationPath: [SendRoute]
 
     @EnvironmentObject private var activityList: ActivityListViewModel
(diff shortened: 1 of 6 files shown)


Bitkit/Views/Wallets/Send/SendPendingScreen.swift
```diff
@@ -89,9 +91,10 @@ struct SendPendingScreen: View {
             } else {
                 app.consumeContactPaymentContext(forPendingPaymentHash: paymentHash)
                 navigationPath.append(.failure(SendFailureContext(
-                    message: t("wallet__payment_failed_description"),
```
jvsena42, line 92: now `message` has no call sites, could delete it

Bitkit/Views/Wallets/Send/SendSheet.swift
```diff
@@ -63,7 +63,7 @@ enum SendRoute: Hashable {
     case tag
     case quickpay
     case pin
-    case pending(paymentHash: String, retryRoute: SendRetryRoute)
+    case pending(paymentHash: String, retryRoute: SendRetryRoute, paymentRequest: String? = nil)
```
jvsena42: all callers pass `paymentRequest` explicitly, is it necessary a default `nil`?

jvsena42 (dismissed):
Tested on emulator, no issues found, just suggestions

Bitkit/Views/Wallets/Send/SendSheet.swift
pwltr: Fixed in [312545f](https://github.com/synonymdev/bitkit-ios/pull/663/commits/312545fc419891bc4f42b0fe2c74162ea26e7661)

Bitkit/Views/Wallets/Send/SendPendingScreen.swift
pwltr: Fixed in [312545f](https://github.com/synonymdev/bitkit-ios/pull/663/commits/312545fc419891bc4f42b0fe2c74162ea26e7661)

jvsena42 (approved):

PR #664, opened 2026-08-12
fix: stop auto-accepting stale monitors

ben-kaufman:
### Description

This PR restores fail-closed Lightning startup when stored channel monitors are stale relative to the channel manager.

- Removes the automatic `DangerousValue` retry that enabled stale-monitor recovery.
- Releases the cross-process Lightning lock when setup fails before a node is created.

The retry was added in 2.1.2 to recover wallets affected by the 2.1.0 RN monitor overwrite. Current ldk-node versions protect newer stored monitors from stale migration data, and those affected wallets are assumed recovered. A future storage mismatch now fails node construction before startup or chain processing. Stale-monitor acceptance stays at the ldk-node default (`false`).

Related: https://github.com/synonymdev/audit/issues/55
Companion: https://github.com/synonymdev/bitkit-android/pull/1155

### Linked Issues/Tasks

- https://github.com/synonymdev/audit/issues/55

### Screenshot / Video

N/A

#### Manual Tests

- [ ] **1.** `regression:` Launch an existing wallet with channels: Lightning node starts normally and balances/channels appear.
- [ ] **2.** `regression:` Force-quit and relaunch: node starts without a stale-monitor recovery log.
- [ ] **3.** Start with a monitor older than the channel manager: startup fails with `DangerousValue`, no node starts, and the foreground Lightning lock is released.

#### Automated Checks

- SwiftFormat and Swift parser checks pass for `LightningService.swift`.
- CI: standard build, unit, integration, and E2E checks run by the PR bot.


Bitkit/Services/LightningService.swift
```diff
@@ -154,46 +154,19 @@ class LightningService {
...
-                } else {
-                    self.node = try builder.buildWithVssStoreAndFixedHeaders(
-                        vssUrl: vssUrl,
-                        storeId: storeId,
-                        fixedHeaders: [:]
-                    )
-                }
-                Logger.info("Stale monitor recovery: build succeeded with accept_stale", context: "Recovery")
             }
         }
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in 108889ff.

`setup` now releases the Lightning lock on any failure before a node is built, and keeps it only after the build succeeds.

Bitkit/Services/LightningService.swift
jvsena42, line 1: nit: could use the opportunity to remove `fetchOrphanedChannelMonitorsIfNeeded` dead code

Bitkit/Services/LightningService.swift
```diff
@@ -75,6 +75,12 @@ class LightningService {
 
         Logger.debug("Checking lightning process lock...")
         try await StateLocker.lock(.lightning, wait: 30) // Wait 30 seconds to lock because maybe extension is still running
+        var shouldReleaseLightningLock = true
```
jvsena42, line 79: nit: could add a comment explaining why this is necessary, so we can remember in the future

jvsena42 (dismissed):
Self-contained PR, migration tests passing https://github.com/synonymdev/bitkit-ios/actions/runs/31698490214

Bitkit/Services/LightningService.swift
ben-kaufman: Added a comment explaining why we release the lock if setup fails before a node exists.

Bitkit/Services/LightningService.swift
ben-kaufman: That's leftover RN channel recovery. It would pull old Lightning channel files from the React Native backup before LDK starts. Nothing calls it anymore so it doesn't run. Leaving it out of this PR.

jvsena42 (approved):

PR #665, opened 2026-08-12
fix: fail closed without vss auth

ben-kaufman:
This PR fails closed when LNURL-auth is missing so wallet backups cannot fall back to unauthenticated VSS.

### Description

Removes the unauthenticated VSS client fallback. If LNURL-auth is not configured, backup setup and Lightning node build now error instead of using a public encryption key or fixed-header VSS.

Env currently always sets an LNURL-auth URL, so this is a footgun fix rather than a user-visible change on production configs.

Related: [synonymdev/audit#55](https://github.com/synonymdev/audit/issues/55) (BITKIT-SEC-003). Library-side rejection is in synonymdev/vss-rust-client-ffi.

### Linked Issues/Tasks

- https://github.com/synonymdev/audit/issues/55

### Screenshot / Video

N/A

#### Manual Tests
- [ ] **1.** Fresh wallet → backup/Lightning start with current Env LNURL-auth URL: setup still succeeds.
- [ ] **2.** `regression:` empty LNURL-auth URL → VSS setup and node build fail instead of using unauthenticated VSS.

#### Automated Checks
- Added `ChannelMigrationPersistenceTests` coverage for retaining migration data on setup failure, clearing it after success, and preserving a newer pending migration.
- SwiftFormat and translation validation pass locally.

Made with [Cursor](https://cursor.com)


Bitkit/Services/LightningService.swift
```diff
@@ -142,7 +142,7 @@ class LightningService {
         let storeId = try await VssStoreIdProvider.shared.getVssStoreId(walletIndex: walletIndex)
 
         let vssUrl = Env.vssServerUrl
-        let lnurlAuthServerUrl = Env.lnurlAuthServerUrl
+        let lnurlAuthServerUrl = Env.lnurlAuthServerUrl.trimmingCharacters(in: .whitespacesAndNewlines)
```
jvsena42, line 152: nit: could trim it once in `Env`

jvsena42 (dismissed):
Self-contained PR, no issues found

Bitkit/Services/VssBackupClient.swift
jvsena42: > nil otherwise.

Stale comment

Bitkit/Services/LightningService.swift
ben-kaufman: These URLs are hardcoded in Env, no user input. Trimming at the two call sites is enough.

Bitkit/Services/VssBackupClient.swift
ben-kaufman: Updated. Nil now means auth is missing and setup fails closed.

jvsena42 (dismissed):

ovitrif:
Follow-up that also consumes vss-rust-client-ffi 0.5.23: https://github.com/synonymdev/bitkit-ios/pull/673

#673 includes the fail-closed VSS path from this PR plus the library bump from synonymdev/vss-rust-client-ffi#20.

The pending-channel-migration persistence from this PR is **not** in #673. If we still want that, it should land separately (or be rebased onto #673).

ovitrif (approved):
Fail-closed LNURL-auth is enforced on backup and Lightning setup, and pending channel-migration data is retained if setup fails.

PR #668, opened 2026-08-13
fix: harden shop payment bridge

ben-kaufman:
### Description

Hardens the Bitrefill shop payment bridge while preserving the intentional QuickPay behavior shared with Android.

1. Accept native shop messages only from the main frame at the exact `https://embed.bitrefill.com` origin and default HTTPS port.
2. Make bridge injection idempotent and retain broader HTTPS Bitrefill navigation separately from the privileged payment origin.
3. Block off-origin checkout navigation with user feedback without restricting the BTC Map caller.
4. Accept only payment request scanner types from shop messages; setup, auth, channel, node, and gift payloads are rejected without clearing existing payment state.
5. Keep QuickPay eligibility independent of payment-PIN settings, matching the accepted product behavior documented on Android PR 1158.

Companion Android PR: https://github.com/synonymdev/bitkit-android/pull/1158

### Linked Issues/Tasks

VulnHunter 01b finding 02 (shop WebView any-origin `payment_intent`). Finding 03 is accepted product behavior and is not changed here.

### Screenshot / Video

N/A, security behavior only.

### QA Notes

- [ ] Shop → gift card → Bitrefill checkout → `payment_intent`: the appropriate Send flow opens with the invoice.
- [ ] Shop checkout stays on HTTPS Bitrefill pages; an off-origin main-frame or window navigation is blocked with a warning.
- [ ] A subframe or non-`embed.bitrefill.com` sender cannot invoke the native payment bridge.
- [ ] A trusted shop message containing a non-payment scanner payload is rejected.
- [ ] QuickPay remains available for an eligible small payment when enabled, including when payment-PIN settings are on.
- [ ] Shop Discover → BTC Map continues to load.

Focused unit coverage: `ShopOriginTests`, `ShopPaymentRequestTests`, and `PaymentNavigationHelperTests`. Translation validation passes.


ovitrif (dismissed):
Looks good. Shop checkout rejects payment_intent messages from non-Bitrefill origins, and QuickPay no longer bypasses PIN-for-payments.

Bitkit/Components/ShopWebView.swift
```diff
@@ -55,28 +55,41 @@ struct ShopWebView: UIViewRepresentable {
...
         ) {
-            decisionHandler(.allow)
+            if navigationAction.targetFrame?.isMainFrame == false {
+                decisionHandler(.allow)
+                return
+            }
+            if ShopOrigin.isAllowed(navigationAction.request.url) {
+                decisionHandler(.allow)
+                return
+            }
```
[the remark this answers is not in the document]
ovitrif: fixed in a23456a

ovitrif:
Simulator QA on `to.bitkit` @ 9c215ea (iPhone 16, iOS 18.5). PIN `1234`, QuickPay on, spending 97316 sats after a Blocktank transfer.

- **1.** Shop → Gift Cards → Amazon.fr €100 cart → Checkout (email step). Did not submit. Shop talks to real `embed.bitrefill.com`, so checkout would emit a mainnet invoice a regtest wallet cannot act on.
- **2.** Gift Cards → eSIMs, product, checkout, and Help stayed on Bitrefill. Privacy Policy stayed on Bitrefill (Bitrefill 404 page). Did not find a live off-Bitrefill main-frame URL to block.
- **3.** PIN on + PIN for payments on + QuickPay on → 500-sat LN invoice: Confirm opened, not QuickPay. Swipe To Pay asked for PIN.
- **4.** PIN on + PIN for payments off + QuickPay on → 600-sat LN invoice: QuickPay opened. The payment itself failed (`Failed to send the given payment`) because the invoice was on a different regtest than this wallet; the check was that QuickPay opened, not that it settled.

Bitkit/Components/ShopWebView.swift
```diff
@@ -55,28 +55,44 @@ struct ShopWebView: UIViewRepresentable {
...
 
         func webView(
             _ webView: WKWebView,
             decidePolicyFor navigationAction: WKNavigationAction,
             decisionHandler: @escaping (WKNavigationActionPolicy) -> Void
         ) {
-            decisionHandler(.allow)
+            if navigationAction.targetFrame?.isMainFrame == false {
+                decisionHandler(.allow)
+                return
```
jvsena42, line 90: `userContentController` only check the top level url frame  `message.webView?.url` not the frame that sends the message. Also `if navigationAction.targetFrame?.isMainFrame == false {` unconditionally allows sub-frame navigation to any origin. So an `iframe` pointed at an attacker origin can call `postMessage` on the handler directly, bypassing the injected event.origin check in `ShopOrigin.messageBridgeScript`; the native guard sees the main frame's `embed.bitrefill.com` URL and forwards to `app.handleScannedData(paymentUri)` -> payment sheet with an attacker-chosen invoice.


should gate `message.frameInfo.securityOrigin.host`

jvsena42 (dismissed):

Bitkit/Components/ShopWebView.swift
ben-kaufman: Fixed in a1a2fa87. The native handler now requires message.frameInfo.isMainFrame and validates message.frameInfo.securityOrigin against the exact https://embed.bitrefill.com origin and default HTTPS port before forwarding. ShopOriginTests covers subframes, alternate Bitrefill hosts, non-HTTPS origins, and non-default ports.

ovitrif (approved):
Approved. Shop checkout now accepts only main-frame payment requests from the trusted Bitrefill embed origin, and QuickPay remains independent of payment-PIN settings.

jvsena42 (approved):

PR #669, opened 2026-08-14
fix: use entered regtest deposit address

pwltr:
## Summary

Fixes the Blocktank regtest deposit form so deposits are sent to the address shown in the address field instead of silently generating a new address for the current primary address type.

## Details

- Uses the visible/editable `depositAddress` value when calling `regtestDepositFunds`
- Trims whitespace/newlines before submitting
- Disables the Deposit button when the address field is empty
- Keeps the normalized submitted address in the field after a successful deposit

## Verification

- Ran `swiftformat` on `Bitkit/Views/Settings/BlocktankRegtestView.swift`
- Verified arm64 simulator build succeeds:
  `xcodebuild -workspace Bitkit.xcodeproj/project.xcworkspace -scheme Bitkit -configuration Debug -sdk iphonesimulator -destination 'generic/platform=iOS Simulator' ARCHS=arm64 ONLY_ACTIVE_ARCH=YES build`

### DIFF
--- a/Bitkit/Views/Settings/BlocktankRegtestView.swift
+++ b/Bitkit/Views/Settings/BlocktankRegtestView.swift
@@ -9,6 +9,7 @@ struct BlocktankRegtestScreen: View {
     @State private var selectedMineBlockCount: Int = 1
     @State private var depositAmount: String = "100000"
     @State private var depositAddress: String = ""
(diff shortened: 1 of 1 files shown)


jvsena42 (approved):
Tested with Segwit and Taproot addresses

PR #672, opened 2026-08-18
fix: add quickpay daily spend limit

### LINKED ISSUE #670: feat: Add QuickPay daily spend limit
## Summary

Port the QuickPay **daily spend limit** behavior from Android [bitkit-android#1159](https://github.com/synonymdev/bitkit-android/pull/1159) to iOS.

Android deliberately keeps QuickPay **PIN-free under limits** (that is the point of QuickPay) and bounds auto-pay with a configurable daily spend multiplier instead of hard-skipping QuickPay whenever PIN-for-payments is on.

Related: iOS [#668](https://github.com/synonymdev/bitkit-ios/pull/668) currently skips QuickPay when PIN-for-payments is enabled. Align iOS with the Android product model when porting this.

## Product behavior (match Android)

- QuickPay stays confirmation-free and PIN-free while under:
  - per-tx threshold (`quickPayAmount` USD), and
  - daily cap: `threshold × dailyLimitMultiplier` USD
- Daily multiplier steps: `1`, `3`, `5`, `10`, `50` (default **`5`**)
- Spend resets on the **local calendar day**
- Record QuickPay USD spend on Success/Pending only (not Error)
- Over the daily cap → Confirm (payment PIN only if that setting is on)
- Keep existing lock/auth deferral behavior as already implemented on iOS

```mermaid
flowchart TD
  incoming[Payment request]
  locked{App unlocked?}
  qpOn{QuickPay on and amount at or under per-tx threshold?}
  daily{spentTodayUsd plus amountUsd at or under dailyCapUsd?}
  quickpay[QuickPay auto-pay no PIN]
  confirm[Confirm then payment PIN if enabled]
  queue[Defer until unlock]

  incoming --> locked
  locked -->|no| queue
  queue --> locked
  locked -->|yes| qpOn
  qpOn -->|no| confirm
  qpOn -->|yes| daily
  daily -->|yes| quickpay
  daily -->|no| confirm
```

## UI

- QuickPay settings: second step slider for the daily multiplier (`1×` … `50×`)
- Copy should show the resolved daily `$` cap from threshold × multiplier

## Reference

- Android PR: https://github.com/synonymdev/bitkit-android/pull/1159
- Android base (lock deferral / shop auth): https://github.com/synonymdev/bitkit-android/pull/1158

## Acceptance

- [ ] Settings expose daily multiplier with the same steps/default as Android
- [ ] Under per-tx + daily cap, QuickPay runs without payment PIN
- [ ] Over daily cap → Confirm; payment PIN honored if enabled
- [ ] Spend is day-keyed and recorded only on QuickPay Success/Pending

ovitrif:
Closes #670

This PR ports QuickPay daily spend limits from Android (synonymdev/bitkit-android#1159) so auto-pay stays PIN-free under a configurable daily cap.

### Description

QuickPay stays confirmation-free and PIN-free while a payment is under the per-tx threshold and the daily cap (`threshold × multiplier` USD). Once the cap would be exceeded, the payment opens Confirm, where PIN for payments still applies if that setting is on.

- Daily multiplier steps: `1`, `3`, `5`, `10`, `50` (default `5`)
- Spend is reserved in USD cents before pay, kept on Success/Pending, and released on failure only from the matching calendar day
- Spend resets on the local calendar day and is kept if the clock rolls back
- Settings add a second slider and show the resolved daily `$` cap
- The multiplier is included in settings backup as `quickPayDailyLimitMultiplier`
- Slider labels and the currency-conversion error are localized. Over-cap is silent Confirm
- Successful QuickPay shows amount plus the paid routing fee
- Over-cap or a lost reserve race replaces QuickPay with Confirm, so Back does not return to the auto-pay loader
- Try Again remounts QuickPay so a new payment starts

### Preview

| QuickPay Settings | Over Cap → Confirm + PIN | 
| - | - |
| <video src="https://github.com/user-attachments/assets/42a655e1-21cb-4a84-bbdb-1e443a201263"> | <video src="https://github.com/user-attachments/assets/f4ac9fd8-90f7-4446-aa85-c4ae144d849a"> |

| Screenshot |
| - |
|  |

#### Manual Tests
- [x] **1.** Settings → QuickPay → enable QuickPay → move threshold and daily multiplier sliders: resolved daily `$` copy updates (`threshold × multiplier`).
- [x] **2.** Unlocked, under per-tx and daily cap, PIN for payments on → scan a fixed-amount LN invoice: QuickPay runs with no payment PIN.
- [x] **3.** Exhaust or exceed the daily cap → scan another eligible invoice: Confirm opens; payment PIN is required if that setting is on.
- [x] **4.** `regression:` Settings → QuickPay → leave QuickPay off → scan a fixed-amount LN invoice: Confirm still opens.
- [x] **5.** `regression:` Variable amount LN invoice / LNURL-pay: still lands on Amount, not QuickPay.
- [x] **6.** QuickPay success: displayed amount includes the paid routing fee.
- [x] **7.** Over daily cap → Confirm: Back does not return to the QuickPay loader.
- [x] **8.** QuickPay failure → Try Again: a new payment starts.

#### Automated Checks
- Unit tests added: day-keyed USD-cent spend reserve/release/reset and clock-rollback keep-spend in `BitkitTests/QuickPaySpendStoreTests.swift`.
- Unit tests added: QuickPay routing under the cap, at the cap, and over the cap, plus replace-QuickPay navigation, in `BitkitTests/PaymentNavigationHelperTests.swift`.
- Unit tests added: fee-inclusive success amount and multiplier fallback in `BitkitTests/QuickPayLimitsTests.swift`.
- Unit tests modified: backup/restore of the daily multiplier, including the Android key, in `BitkitTests/AddressTypeSettingsTests.swift`.
- `node scripts/validate-translations.js`: 0 errors.
- Focused unit tests passed locally on iPhone 16 (iOS 18.5).


Bitkit/Views/Wallets/Send/SendQuickpay.swift
```diff
@@ -87,6 +103,34 @@ struct SendQuickpay: View {
...
+        else {
+            throw AppError(
+                message: t("wallet__send_quickpay__currency_conversion"),
+                debugMessage: "Currency conversion failed"
+            )
+        }
+
+        let dayKey = QuickPaySpendStore.dayKey()
+        let reserved = spendStore.tryReserve(amountUsd: amountUsd, dayKey: dayKey, dailyCapUsd: dailyCapUsd)
+        guard reserved else {
```
ben-kaufman: `tryReserve` can still fail after routing decides QuickPay is eligible, for example if another reservation lands first. This currently opens the generic failure screen and retry goes straight back to QuickPay. Could we route this case to Confirm so the payment follows the documented daily-limit and PIN behavior?

Bitkit/Views/Wallets/Send/SendQuickpay.swift
```diff
@@ -59,26 +62,39 @@ struct SendQuickpay: View {
...
+            do {
+                let settled = try await wallet.sendWithTimeout(
+                    bolt11: bolt11,
+                    sats: nil,
+                    onTimeout: {
+                        app.addPendingPaymentHash(paymentHash)
+                        navigationPath.append(.pending(paymentHash: paymentHash, retryRoute: .quickpay, paymentRequest: bolt11))
+                    }
+                )
+                wallet.sendAmountSats = QuickPayLimits.amountWithFeeSats(
```
ben-kaufman: This fee update only runs when the payment settles before the timeout. If it succeeds from Pending, `AppViewModel` drops the fee and Success still shows the original amount. Could we carry the fee through the pending resolution and apply the same amount update there?

ben-kaufman (commented):
Two findings below.

Bitkit/Views/Wallets/Send/SendQuickpay.swift
```diff
@@ -59,26 +62,39 @@ struct SendQuickpay: View {
...
+                    amountSats: amountSats,
+                    feePaidSats: settled.feePaidSats
+                )
+                Logger.info("Quickpay payment successful: \(paymentHash)")
+                navigationPath.append(.success(paymentId: paymentHash))
+            } catch is PaymentTimeoutError {
+                // Pending keeps the reserved spend; onTimeout already navigated.
+                return
```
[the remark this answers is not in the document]
ovitrif: Timed-out QuickPay payments now keep a hash-keyed reservation. A later terminal failure releases it so a retry cannot consume the daily cap twice; success still keeps the spend.

Bitkit/Views/Wallets/Send/SendQuickpay.swift
ovitrif: If `tryReserve` loses the race after QuickPay is already open, we now continue on Confirm (or LNURL Confirm) instead of the generic failure screen, so payment PIN still applies.

Bitkit/Views/Wallets/Send/SendQuickpay.swift
ovitrif: Pending success now carries `feePaidSats` through `SendSheetPendingResolution` and applies the same amount-plus-fee update before Success.

Bitkit/Views/Wallets/Send/SendQuickpay.swift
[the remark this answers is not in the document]
ben-kaufman: `trackPending` still runs too late. The task group cannot finish throwing the timeout until the continuation-based send receives its terminal event, so `AppViewModel` calls `releasePending` or `forgetPending` before this record exists. Could we track the reservation inside `onTimeout` before adding the pending hash and route?

Bitkit/Utilities/QuickPaySpendStore.swift
```diff
@@ -0,0 +1,147 @@
...
+        guard !paymentHash.isEmpty else { return }
+
+        lock.lock()
+        defer { lock.unlock() }
+
+        var pending = lockedPendingReservations()
+        guard let reservation = pending.removeValue(forKey: paymentHash) else { return }
+        lockedWritePending(pending)
+
+        let spend = lockedSpend(forDayKey: reservation.dayKey)
```
ben-kaufman: An old pending failure can subtract from the new day's spend. Once the store advances to Aug 16, releasing an Aug 15 reservation goes through the clock rollback branch and decrements the Aug 16 bucket. Could we store the effective accounting day with each reservation and only release from a matching bucket?

Bitkit/Views/Wallets/Send/SendPendingScreen.swift
```diff
@@ -85,6 +85,12 @@ struct SendPendingScreen: View {
             app.consumeSendSheetPendingResolution(paymentHash: paymentHash)
             if resolution.success {
                 Task { @MainActor in
+                    if let feePaidSats = resolution.feePaidSats, let amountSats = wallet.sendAmountSats {
```
ben-kaufman: This now adds routing fees for every pending Lightning send, not just QuickPay. Normal Confirm payments still show the invoice amount when they settle immediately, so the displayed amount changes based on timing. Could we gate this update to `retryRoute == .quickpay`, or update the immediate Confirm paths too?

Bitkit/Views/Wallets/Send/SendQuickpay.swift
```diff
@@ -87,6 +112,30 @@ struct SendQuickpay: View {
...
+                message: t("wallet__send_quickpay__currency_conversion"),
+                debugMessage: "Currency conversion failed"
+            )
+        }
+
+        let dayKey = QuickPaySpendStore.dayKey()
+        let reserved = spendStore.tryReserve(amountSats: amountSats, dayKey: dayKey, dailyCapSats: dailyCapSats)
+        guard reserved else {
+            Logger.info("Skipping QuickPay pay: daily spend reserve failed for '\(amountSats)'")
+            navigationPath.append(PaymentNavigationHelper.confirmRouteAfterQuickPayCap(app: app))
```
ben-kaufman: This pushes Confirm on top of the QuickPay root. Confirm shows Back, which returns to the same QuickPay view with `didStartPayment == true`, so it stays on an inactive loader. Could we replace QuickPay in the flow here, or prevent Back from returning to it?

ben-kaufman (commented):
Three new findings from the re-audit.

ovitrif:
Drafted to fix the issues surfaced in the last review round.

Bitkit/Views/Wallets/Send/SendQuickpay.swift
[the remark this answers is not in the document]
ovitrif: Reservation is bound after the event listener is live and before pending is tracked. A later terminal failure releases that hash-keyed reservation so a retry cannot consume the cap twice.

Bitkit/Utilities/QuickPaySpendStore.swift
ovitrif: Release now only decrements when the reservation day matches the stored accounting day. An Aug 15 pending failure no longer subtracts from Aug 16 spend.

Bitkit/Views/Wallets/Send/SendPendingScreen.swift
ovitrif: Pending success only adds the routing fee when `retryRoute == .quickpay`. Confirm pending still shows the invoice amount.

Bitkit/Views/Wallets/Send/SendQuickpay.swift
ovitrif: Over-cap now replaces QuickPay in the stack (and clears QuickPay ownership first). Back no longer returns to the inactive QuickPay loader.

Bitkit/Views/Wallets/Send/SendQuickpay.swift
```diff
@@ -34,19 +39,21 @@ struct SendQuickpay: View {
         .sheetBackground()
         .frame(maxWidth: .infinity, maxHeight: .infinity)
         .onAppear {
+            guard !didStartPayment else { return }
```
ovitrif: **[bug]** Try Again from a QuickPay failure (immediate fail or pending-then-fail) does not start a new payment. `resetNavigationForRetry` calls `resetQuickPay()` then pops back to the existing `.quickpay` root/destination (`navigationPath = []` or `[.quickpay]`). That reuses the same `SendQuickpay` instance: `didStartPayment` stays true, and popping a `NavigationStack` destination does not reliably re-run `onAppear` on the remaining view. Android retry pops the send graph and navigates to a fresh QuickPay screen, so `pay()` runs again. iOS leaves the user on the inactive loader. `resetQuickPay()` on retry only helps if `performPayment()` actually runs.

**Suggestion:** Give `SendQuickpay` a new identity on QuickPay retry, same as Android recreating the screen. In `SendSheet`, keep a `quickPaySession` counter, apply `.id(quickPaySession)` to the `SendQuickpay` view, and increment the counter next to `app.resetQuickPay()` in `resetNavigationForRetry`. Keep the `didStartPayment` / `beginQuickPay()` guards for same-instance `onAppear` re-entry.

Bitkit/Utilities/QuickPayLimits.swift
```diff
@@ -0,0 +1,51 @@
...
+    static func thresholdCents(_ thresholdUsd: Double) -> Int64 {
+        Int64(Int(thresholdUsd)) * 100
+    }
+
+    static func capCents(thresholdUsd: Double, multiplier: Double) -> Int64 {
+        thresholdCents(thresholdUsd) * Int64(Int(sanitizedMultiplier(multiplier)))
+    }
+
+    static func reserveCents(convertedCents: Int64, thresholdUsd: Double) -> Int64 {
+        min(convertedCents, thresholdCents(thresholdUsd))
```
piotr-iohk: Same hole as Android: `usdCents` rounds sub-cent invoices (e.g. `7` sats → `$0.00`) to `0`, so `reserveCents` is `0`. `spent + 0 <= cap` still passes when the day is already full, and dust from a zero ledger never counts. Could we treat any positive sat amount as at least `1` cent in `reserveCents` and cover a `0`-cent conversion at a full cap in `QuickPayLimitsTests` / `QuickPaySpendStoreTests`?

pwltr:
Tested the QuickPay daily limit flow on local regtest with fixed-amount hold invoices.

Covered:
- QuickPay under the per-payment threshold and daily cap stays PIN-free and moves to Pending after the hold invoice timeout.
- Settling the hold invoice moves Pending -> Success.
- Canceling the hold invoice moves Pending -> Failure.
- Exhausting the daily cap routes the next eligible fixed invoice to Confirm without returning to the QuickPay loader on Back.
- Failed QuickPay reservations are released, while successful/pending spend remains counted against the daily cap.
- Try Again from QuickPay failure starts a fresh payment attempt.
- QuickPay-off and variable-amount invoice routes still go through the expected non-QuickPay screens.

The timeout now starts after `lightningService.send` returns the payment hash. That seems fine if LDK submission is expected to return quickly, but it’s worth keeping in mind that pre-hash submission time is no longer covered by the pending-screen timeout.

I also reviewed the code structure around `QuickPaySpendStore`, `QuickPayLimits`, and the send navigation changes. The implementation is reasonably contained and the focused unit tests cover the important ledger/day-boundary behavior. No blocking issues from my side.

piotr-iohk (dismissed):
Tested on emulator. Tried more than `$1 × 1` — also `$50 × 1`, `$1 × 50`, and other slider combos.

**Passed**
- Settings: both sliders and the resolved `$` copy (`threshold × multiplier`)
- Under-cap QuickPay with PIN for payments on — no Confirm, no payment PIN
- Filling the cap with normal amounts → Confirm + PIN
- Cold start / locked with PIN — scan stays deferred until unlock, then follows the cap rules
- Raising the multiplier with the app unlocked does not ask for payment PIN
- Over-cap Confirm → Back does not return to a dead QuickPay loader
- QuickPay failure → Try Again starts a new payment; next under-cap invoice still QuickPays
- Normal Confirm (QuickPay off, or over the per-tx threshold) — success amount is the invoice amount, including via Pending
- Variable-amount LN / LNURL-pay still lands on Amount
- QuickPay off still opens Confirm
- Wipe + restore: multiplier comes back; spend does not (fresh allowance)

**Failed**
1. **Dust invoices skip the daily cap.** Amounts that display `$0.00` (I used `7` sats) reserve `0` cents, so `spent + 0 <= cap` still passes when the day is already full, and a streak of those invoices never increments the ledger. Same hole on Android. See https://github.com/synonymdev/bitkit-ios/pull/672#discussion_r3822536777

Wipe + restore is the other way on Android (spend comes back with metadata backup). Called out on https://github.com/synonymdev/bitkit-android/pull/1159 — they need to agree which behavior both apps use.

Requesting changes for the dust issue.

Bitkit/Utilities/QuickPayLimits.swift
ovitrif: Same rounding as Android. `usdCents` rounds `value * 100` with `.plain`, so a few sats (your `7` sat example) become `$0.00` / `0` cents. `canApply` and `tryReserve` both use `reserveCents`.

Leaving this as-is here too: https://github.com/synonymdev/bitkit-android/pull/1159#discussion_r3822893538

ben-kaufman (commented):
Two findings from the latest re-audit.

Bitkit/Utilities/QuickPaySpendStore.swift
```diff
@@ -0,0 +1,226 @@
...
+        return false
+    }
+
+    func tryReserve(
+        amountSats: UInt64,
+        thresholdUsd: Double,
+        multiplier: Double,
+        rates: QuickPaySpendRates
+    ) throws -> QuickPaySpendReservation? {
+        guard let convertedCents = rates.satsToUsdCents(amountSats) else {
```
ben-kaufman, line 231: `tryReserve` does not recheck the current per-payment threshold. If FX rates refresh after routing selects QuickPay, an invoice can cross the threshold but still auto-pay because `reserveCents` clamps it back to the threshold. Could we verify `amountSats <= rates.usdToSats(thresholdUsd)` here and fall back to Confirm when it no longer qualifies?

Bitkit/ViewModels/AppViewModel.swift
```diff
@@ -1054,6 +1085,10 @@ extension AppViewModel {
         case let .paymentFailed(paymentId, paymentHash, reason):
             let hash = paymentId ?? paymentHash
             if let hash, pendingPaymentHashes.contains(hash) {
```
ben-kaufman: QuickPay reservations persist across relaunch, but `pendingPaymentHashes` does not. If LDK delivers a failure after restart, this guard skips `release` and the failed payment keeps consuming today's cap. Could we resolve matching store reservations before the UI-only pending-hash check and cover the relaunch case?

ovitrif:
> Wipe + restore is the other way on Android (spend comes back with metadata backup). Called out on [synonymdev/bitkit-android#1159](https://github.com/synonymdev/bitkit-android/pull/1159) — they need to agree which behavior both apps use.

Wipe + restore now keeps today’s spend. Metadata backup includes quickPaySpendDayKey, quickPaySpentCentsToday, and quickPayReservations on AppCacheData, same keys as Android. Restore writes them back through QuickPaySpendStore.

Dust is unchanged: https://github.com/synonymdev/bitkit-ios/pull/672#discussion_r3822912256

Bitkit/Models/SettingsBackupConfig.swift
```diff
@@ -65,6 +69,7 @@ enum SettingsBackupConfig {
         "warnWhenSendingOver100": "enableSendAmountWarning",
         "bitcoinDisplayUnit": "displayUnit",
         "enableQuickpay": "isQuickPayEnabled",
+        "quickpayDailyLimitMultiplier": "quickPayDailyLimitMultiplier",
```
ben-kaufman, line 71: The cap’s two settings do not round-trip together. iOS writes `quickpayAmount`, while Android PR #1159 reads `quickPayAmount`, so Android ignores the configured threshold and falls back to $5. For example, a $1 × 50 limit becomes $5 × 50 after restore. Could we also map `quickpayAmount` to `quickPayAmount` and cover both restore directions?

Bitkit/Models/BackupPayloads.swift
```diff
@@ -104,6 +113,9 @@ struct AppCacheData: Codable {
         highBalanceIgnoreTimestamp = try c.decode(TimeInterval.self, forKey: .highBalanceIgnoreTimestamp)
         dismissedSuggestions = try c.decode([String].self, forKey: .dismissedSuggestions)
         lastUsedTags = try c.decode([String].self, forKey: .lastUsedTags)
+        quickPaySpendDayKey = try c.decodeIfPresent(String.self, forKey: .quickPaySpendDayKey) ?? ""
```
ben-kaufman: An Android #1159 metadata payload lacks the iOS-specific cache fields required above, so decoding throws before reaching these shared QuickPay fields. `performRestore` then skips the metadata and marks it synced, restoring the ledger as empty. Could we default the platform-specific cache fields or decode the shared QuickPay state separately, with an Android metadata fixture test?

Bitkit/Utilities/QuickPayLimits.swift
ovitrif: Positive sat amounts that convert to `$0.00` now reserve `1` cent in `reserveCents` (`canApply` and `tryReserve`). A full cap no longer QuickPays that dust.

Bitkit/Utilities/QuickPaySpendStore.swift
ovitrif: `tryReserve` now rechecks `usdToSats(thresholdUsd)` and returns nil when the invoice is over the live sat threshold, so the payment continues on Confirm.

Bitkit/ViewModels/AppViewModel.swift
ovitrif: `release` / `clear` now run on the LDK terminal event even if `pendingPaymentHashes` is empty. Toast and pending navigation stay behind that set.

Bitkit/Models/SettingsBackupConfig.swift
ovitrif: `quickpayAmount` now maps to Android `SettingsData.quickPayAmount`, same adapter as the multiplier. Backup writes `quickPayAmount`; restore reads that key. A `$1` threshold stays `$1`.

Bitkit/Models/BackupPayloads.swift
ovitrif: iOS-only cache fields are optional. Android `AppCacheData` JSON without intro keys decodes; spend/reservations restore; missing intro keys are not written as `false`.

ovitrif:
drafted to fix the bot review point

Bitkit/ViewModels/WalletViewModel.swift
```diff
@@ -869,13 +874,18 @@ class WalletViewModel: ObservableObject {
...
-    ) async throws -> PaymentHash {
-        try await withThrowingTaskGroup(of: PaymentHash.self) { group in
-            group.addTask { try await self.send(bolt11: bolt11, sats: sats) }
+        afterListening: (@MainActor (String) -> Void)? = nil,
+        onTimeout: (@MainActor (String) -> Void)? = nil
+    ) async throws -> SettledLightningPayment {
+        let hash = try await lightningService.send(bolt11: bolt11, sats: sats)
+        let paymentHash = String(hash)
+        return try await withThrowingTaskGroup(of: SettledLightningPayment.self) { group in
+            group.addTask { try await self.watchSend(hash: paymentHash, afterListening: afterListening) }
```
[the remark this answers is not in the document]
ovitrif: resolved in `8d390f1`

BitkitTests/QuickPayPaymentCoordinatorTests.swift
```diff
@@ -0,0 +1,750 @@
...
+        )
+        await fulfillment(of: [sendStarted], timeout: 2)
+        coordinator.detach()
+
+        coordinator.pay(
+            app: appWithInvoice,
+            wallet: WalletViewModel(),
+            settings: settings,
+            currency: CurrencyViewModel(),
+            presentation: presentation(onRoute: { _ in XCTFail("Live send should not emit yet") }, onConfirm: { didConfirm = true })
```
ben-kaufman: This failure callback can run after the test returns. Once `sendCont` resumes, the coordinator may emit later while another test is running, which is why the current unit job reports this message under `testRegisterPostsDescriptorFormPayload`. Could we detach or cancel the coordinator, or await its final route before returning?

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
```diff
@@ -0,0 +1,521 @@
...
+
+        if store.record(matching: invoiceHash) == nil || op.terminalSuccess == true {
+            applyFee(op, wallet: wallet)
+            emitSuccess(op)
+            return
+        }
+
+        let attached = op.presentation ?? presentation
+        do {
+            let settled = try await wallet.waitForLightningPayment(hash: invoiceHash) { _ in
```
ben-kaufman, line 322: An LDK terminal event can arrive after the checks above but before `watchSend` registers its handler. `complete` then clears the ledger while no pending hash exists, and the new listener misses the one-shot event. The timeout later opens Pending with no event left to resolve it. Could we register before this gap or recheck `op.terminalSuccess` immediately after listener setup?

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
```diff
@@ -0,0 +1,521 @@
...
+            }
+        }
+    }
+
+    static func isHardReject(_ error: Error) -> Bool {
+        guard let nodeError = error as? NodeError else {
+            return false
+        }
+        switch nodeError {
+        case .InvalidInvoice, .InvalidAmount, .InvalidPaymentHash, .InvalidPaymentId, .InvalidNetwork:
```
ben-kaufman, line 165: `NodeError.NotRunning` falls through as an ambiguous error even though the node cannot submit the payment. With no matching row, this keeps the reservation and opens Pending, but no terminal event can arrive to resolve it. Could we treat `NotRunning` and other known pre-dispatch errors as hard rejects?

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
```diff
@@ -0,0 +1,521 @@
...
+    private func removeOp(_ op: InFlightOp) {
+        operations = operations.filter { $0.value !== op }
+    }
+
+    private var uniqueOps: [InFlightOp] {
+        var seen = Set<ObjectIdentifier>()
+        return operations.values.filter { seen.insert(ObjectIdentifier($0)).inserted }
+    }
+
+    private func applyFee(_ op: InFlightOp, wallet: WalletViewModel, feePaidSats: UInt64? = nil) {
```
ben-kaufman, line 481: `detach()` leaves the old run active, and this applies its fee to the shared current `wallet.sendAmountSats`. If payment A is dismissed, the user starts payment B, and A then succeeds, this changes B’s amount to B plus A’s fee. Could we retain A’s original amount on the operation or prevent detached sessions from mutating the current send state?

Bitkit/Models/BackupPayloads.swift
```diff
@@ -104,6 +107,7 @@ struct AppCacheData: Codable {
         highBalanceIgnoreTimestamp = try c.decode(TimeInterval.self, forKey: .highBalanceIgnoreTimestamp)
         dismissedSuggestions = try c.decode([String].self, forKey: .dismissedSuggestions)
         lastUsedTags = try c.decode([String].self, forKey: .lastUsedTags)
+        quickPayLedger = try c.decodeIfPresent(QuickPayLedger.self, forKey: .quickPayLedger)
```
ben-kaufman, line 132: Adding `quickPayLedger` here still does not make the Android payload decodable. Android’s `AppCacheData` omits the iOS intro and timestamp fields, but the required `decode(...)` calls above throw before this field is reached. Could we default the platform-only fields or extract the shared ledger separately, and add an Android-shaped fixture test?

BitkitTests/QuickPayPaymentCoordinatorTests.swift
ovitrif: Thanks. The zombie test was resuming send and returning while the first run was still alive, and the second `pay` had reattached the `XCTFail` presentation onto that op. CI then reported it under `testRegisterPostsDescriptorFormPayload`. I gated the fail to the no-emit window and `detach()` before resume. Same cleanup on `testPayIgnoresReentryUntilDetach`, which had the same resume-and-return shape.

Resolved in [f3fae45](https://github.com/synonymdev/bitkit-ios/commit/f3fae45)

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
ovitrif: The leftover window is real. It is the sibling of the `afterListening` race we already patched. Instant success during `send()` is covered. After `send()` returns, `waitForLightningPayment` yields on a task group before `addOnEvent`, so AppViewModel can `complete()` in that yield.

`complete()` does not emit UI on purpose. The ledger is already settled. The user can see a success toast and then sit on Pending. Rechecking `terminalSuccess` after the listener is on is a small follow-up. I'll keep that out of this PR. Happy to track it if you want.

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
ovitrif: The stuck-Pending path is real if `send` throws `NotRunning` with no LDK row: we keep the reservation and no terminal event arrives.

Android's `classifyDispatchError` uses the same list: `InvalidInvoice`, `InvalidAmount`, `InvalidPaymentHash`, `InvalidPaymentId`, `InvalidNetwork`. `NotRunning` is ambiguous there too. Treating it as a hard reject here would diverge. If we change this, it should be both platforms together.

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
ovitrif: This path is real. `detach()` clears presentation but the run keeps going, and `applyFee` writes A's fee onto the shared `wallet.sendAmountSats`. Dismiss A, start B, A succeeds: B's displayed amount becomes B plus A's fee.

That is a display mix-up. The daily cap is already settled. Android keeps `displaySats` on the session and does not mutate a shared amount. Rare enough that I'll leave it out of this PR. Happy to track it if you want.

Bitkit/Models/BackupPayloads.swift
ovitrif: The throw is real, and it now drops the daily-cap ledger too. It was not a visible hard fail: `performRestore` swallowed the error, marked metadata synced, and restore looked fine with no spend.

iOS and Android `AppCacheData` are still different structs. I kept the stored properties required and defaulted missing keys with `decodeIfPresent` in `init(from:)`, same pattern as `hasDismissedWidgetsOnboardingHint`. Real iOS backups still write every key, so they round-trip the same. An Android-shaped metadata fixture now restores the ledger.

That also lets the rest of Android→iOS metadata apply: tags, pubky session, and contact overrides. The required intro fields were throwing before `MetadataBackupV1.cache` was read, so the whole category was skipped.

Resolved in [ddcfd4e](https://github.com/synonymdev/bitkit-ios/commit/ddcfd4e)

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
```diff
@@ -0,0 +1,521 @@
...
+        } catch {
+            fail(presentation, error: error, bolt11: bolt11)
+            return
+        }
+
+        if let existing = operations[invoiceHash] {
+            existing.presentation = presentation
+            if existing.emitted {
+                replayPending(existing)
+            } else if !existing.runActive {
```
ben-kaufman, line 239: If the first QuickPay session is dismissed before its timeout, `emitPending` returns because the presentation is nil while the payment wait remains active. Rescanning the same invoice then reaches this branch with `runActive == true` and emits no route, so QuickPay stays on the spinner until another LDK event. Could we make the timeout unwind cleanly or record and replay Pending even when the first presentation is detached?

ovitrif: The spinner can't outlive the first session's timer. `detach()` nils the presentation but doesn't cancel the run, so the original 10s timeout inside `waitForLightningPayment` stays armed. The rescan reattaches its presentation to the live op, and `emitPending` / `emitSuccess` / `emitFailure` all read `op.presentation` at fire time — so the rescanned session gets Pending (or the terminal route) when the original timer fires, within the remainder of the 10s window. No LDK event needed.

The nil-presentation no-op only consumes the timer when nothing reattached; that path unwinds `runActive = false` on return, and a later rescan then hits the `!runActive` branch and emits Pending immediately. Worst case is the loader for the remaining seconds of the original window — same UX as a fresh QuickPay wait, just shorter. Leaving as-is.

ben-kaufman: `onTimeout` fires, but `waitForLightningPayment` does not return at that point. Its task group still waits for `watchSend`, which is suspended in a checked continuation and does not observe cancellation, so `runActive` stays true until an LDK event. I reproduced detach, wait past timeout, then rescan with a focused test and the second session received no route. Rescanning before the timer fires works. Could we keep this open and handle the post-timeout rescan case?

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
```diff
@@ -0,0 +1,521 @@
...
+        for key in keys {
+            if let op = operations[key] {
+                removeOp(op)
+                return
+            }
+        }
+    }
+
+    static func isHardReject(_ error: Error) -> Bool {
+        guard let nodeError = error as? NodeError else {
```
ben-kaufman: `LightningService.send` wraps LDK failures in `AppError`, so both direct `NodeError` casts here miss production errors. A hard reject then opens Pending and keeps its reservation, while a duplicate rescan can keep a second daily-cap charge. Could we unwrap `AppError.underlyingError` in both classifiers and cover the wrapped cases in tests?

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
ovitrif: You're right — my previous reply had the group semantics wrong. The body throw doesn't propagate until all children finish, and `watchSend`'s continuation only resumed on an LDK event, so the run outlived the timer and `runActive` stayed true.

Fixed in `5c2ba1dc` by making the timeout unwind cleanly. `watchSend` wraps its continuation in `withTaskCancellationHandler`; when the group cancels it after the timer throws, it resumes with `CancellationError`, `waitForLightningPayment` returns `PaymentTimeoutError`, and `runActive` resets. A post-timeout rescan then takes the `!runActive` branch and gets Pending immediately. Rescanning before the timer fires keeps the existing reattach behavior. `WalletViewModelSendTests` covers both unwind paths: timeout with no LDK event, and cancellation of the outer task.

Bitkit/Utilities/QuickPayPaymentCoordinator.swift
ovitrif: Fixed in `72438fe8`. Both classifiers now unwrap `AppError.underlyingError` before the `NodeError` checks, same shape as Android's `asNodeException()`. Wrapped cases are covered in tests: a wrapped `InvalidInvoice` hard-rejects and releases the reservation, and a wrapped `DuplicatePayment` with a succeeded LDK row resolves to success without a second cap charge.

ben-kaufman (dismissed):

piotr-iohk:
Retested. Previous findings are fixed:
- Dust invoices showing $0.00 no longer QuickPay once the daily cap is full (both platforms).
- Wipe+restore: iOS used to reset today’s spend (fresh QuickPay allowance) while Android kept it. iOS now matches Android — multiplier and today’s spend both come back.

New issue, both platforms: paying an invoice via QuickPay, then scanning/pasting that same invoice again, shows **Bitcoin Sent** / success again. LDK rejects it as already paid (`DuplicatePayment` / "an invoice must not be paid twice"). There is no second payment and it does **not** add to the daily cap — the Success screen is a replay (and omits the routing fee).

Recordings + logs attached.

[bitkit_logs_2026-08-27_10-54-19.zip](https://github.com/user-attachments/files/31511781/bitkit_logs_2026-08-27_10-54-19.zip)

https://github.com/user-attachments/assets/3e881837-98ff-4683-8a61-a5264eaf2edf

piotr-iohk:
Nit: iOS shows a **Payment Sent** toast on every QuickPay success, on top of the Bitcoin Sent screen. Android has the same strings but suppresses that toast while the QuickPay sheet is open (`isQuickPayHandling`). 

Screenshot 2026-08-27 at 13 13 47

ovitrif:
Drafted to address review findings.

ovitrif:
Addressed the latest QuickPay findings in `9f5c3eb2`:

- A fresh retry of an already-succeeded invoice now releases only the new reservation and opens Payment Failed instead of replaying Bitcoin Sent. Previously counted daily spend remains unchanged.
- Payment success and failure toasts are suppressed while the matching QuickPay Send sheet is open; background settlement after dismissing the sheet still shows its toast.

Focused `QuickPayPaymentCoordinatorTests`: 29 passed, 0 failed. I added QA items 9 and 10 to the PR description for retesting.

piotr-iohk (approved):
Retested, LGTM.

